PR #144, opened 2015-01-27
- use C++11 chrono::steady_clock

aschnell:


lslezak:
LGTM, nice C++11 feature!

(C++ becomes more and more usable, who would expect that after so many years... :stuck_out_tongue_closed_eyes:)

PR #147, opened 2015-02-16
- added helper to install into subvolume (fate#318392)

aschnell:


client/installation-helper.cc
```diff
@@ -0,0 +1,226 @@
...
+	SysconfigFile config(CONFIGTEMPLATEDIR "/" "default");
+
+	config.setName(tmp_mount.getFullname() + CONFIGSDIR "/" "root");
+
+	config.setValue(KEY_SUBVOLUME, "/");
+	config.setValue(KEY_FSTYPE, "btrfs");
+    }
+    catch (const FileNotFoundException& e)
+    {
+	cerr << "copying config-file failed" << endl;
```
jreidinger: I do not get why it is called copying when it looks like creating new file using template and change some values for me.

client/installation-helper.cc
```diff
@@ -0,0 +1,226 @@
...
+    // step runs in inst-sys
+
+    cout << "step 2 device:" << device << " root-prefix:" << root_prefix
+	 << " default-subvolume-name:" << default_subvolume_name << endl;
+
+    cout << "mounting device" << endl;
+
+    string subvol_option = default_subvolume_name;
+    if (!subvol_option.empty())
+	subvol_option += "/";
```
jreidinger: are you sure condition here is correct? I think in case of empty subvol_option it should use "/". If it is right, then I think it deserve comment why it is this way as it is slightly confusing.

client/installation-helper.cc
```diff
@@ -0,0 +1,226 @@
...
+    subvol_option += ".snapshots";
+
+    SDir s_dir(root_prefix + "/.snapshots");
+    if (!s_dir.mount(device, "btrfs", 0, "subvol=" + subvol_option))
+    {
+	cerr << "mounting .snapshots failed" << endl;
```
jreidinger: can we give user some details why it failed or e.g. output of mount?

client/installation-helper.cc
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    cout << "done" << endl;
```
jreidinger: so in case of error it write mounting snapshots failed and then it write done to cout. I worry it can confuse users, maybe printing error without done will be better.

client/installation-helper.cc
```diff
@@ -0,0 +1,226 @@
...
+    cout << "modifying sysconfig-file" << endl;
+
+    try
+    {
+	SysconfigFile sysconfig(SYSCONFIGFILE);
+	sysconfig.setValue("SNAPPER_CONFIGS", { "root" });
+    }
+    catch (const FileNotFoundException& e)
+    {
+	cerr << "sysconfig-file not found" << endl;
```
jreidinger: so if I get code it is not fatal to not modify sysconfig and it can continue with other settings, right?

client/installation-helper.cc
```diff
@@ -0,0 +1,226 @@
...
+	description = opt->second;
+
+    if (step == "1")
+	step1(device, description);
+    else if (step == "2")
+	step2(device, root_prefix, default_subvolume_name);
+    else if (step == "3")
+	step3(root_prefix, default_subvolume_name);
+    else if (step == "4")
+	step4();
```
jreidinger: If I get it is internal only tool, so error handling like not passing options is not needed?

client/installation-helper.cc
aschnell: The output is only visible in the YaST installation log.

client/installation-helper.cc
aschnell: Yes, internal tool only called from YaST. That's why it is placed in /usr/lib/snapper/...

snapper/Snapshot.cc
```diff
@@ -596,110 +533,82 @@ namespace snapper
...
 	Snapshot snapshot(snapper, POST, nextNumber(), time(NULL));
 	snapshot.pre_num = pre->getNum();
-	snapshot.uid = uid;
-	snapshot.description = description;
-	snapshot.cleanup = cleanup;
-	snapshot.userdata = userdata;
+	snapshot.uid = scd.uid;
+	snapshot.description = scd.description;
+	snapshot.cleanup = scd.cleanup;
+	snapshot.userdata = scd.userdata;
```
jreidinger: I see this at least three times, maybe it make sense to add method to snapshot that get scd and set data inside?
so this code can look like:
```
  snapshot.configScd(scd);
```

snapper/Btrfs.cc
```diff
@@ -1422,19 +1439,27 @@ namespace snapper
 
     private:
 
+	string xxx() const	// TODO
```
jreidinger: maybe call it `target_fstab`?

client/installation-helper.cc
aschnell: Is correct, the subvol mount option is either "@/.snapshots" or just ".snapshots" depending on whether default_subvolume_name is "@" or "".

snapper/AsciiFile.h
```diff
@@ -80,13 +82,14 @@ namespace snapper
...
-	const string Name_C;
-	const bool remove_empty;
-
 	vector<string> Lines_C;
 
-    };
+    private:
 
+	string Name_C;
+	bool remove_empty;
```
jreidinger: I know that it is just copy of already existing, but naming looks quite inconsistent, does it follow any C++ style guide?

client/installation-helper.cc
aschnell: The mount function has logging.

client/installation-helper.cc
jreidinger: OK, thanks, clear explanation, can you please add it to comment there, so it helps others reading code?

client/installation-helper.cc
jreidinger: OK

client/installation-helper.cc
jreidinger: OK, if user do not see it on console, it is fine for me

snapper/Snapshot.cc
aschnell: I doubt it will make the code shorter and more readable.

snapper/AsciiFile.h
aschnell: Maybe it does follow some style guide but not one I follow. It is just old code.

snapper/Snapshot.cc
jreidinger: it will make it easier to modify in future, because if you add another part to scd, you just need to modify one place to pass it to snapshot. Also if you need to do some operation like cleaning a value from scd, then it will be done in one place.
But it is not hard requirement, just idea

snapper/AsciiFile.h
jreidinger: OK

client/installation-helper.cc
aschnell: I think not fatal enough to abort the installation.

snapper/Snapshot.cc
aschnell: But maybe you don't want to set all values in the future.

jreidinger:
LGTM

PR #150, opened 2015-02-24
- drop -mt decoration from boost libraries

aschnell:


aschnell:
I don't increase the snapper version so often because it is much (non automated) work, e.g. I upload the tarball to ftp.suse.com and update the obs project.

For this small change I don't see necessity to do so.

gabi2:
ok, LGTM

PR #154, opened 2015-03-01
Install the included systemd units

GreatEmerald:
Right now the systemd timer units are in the tarball, but aren't actually installed anywhere. This installs them (prefixing "snapper-" to avoid conflicts).


aschnell:
The pull request lacks several things, e.g. adding the files to EXTRA_DIST and installing the systemd-helper.

I have created https://github.com/openSUSE/snapper/pull/155.

aschnell:
OK, EXTRA_DIST was already merged - it just was missing in my checkout.

PR #155, opened 2015-03-02
- install systemd unit files and helper (see gh#openSUSE/snapper#154)

aschnell:


jreidinger:
LGTM, just question, I see new services, but do not see any restart or reload in post-install part of spec, is it correct?

aschnell:
Sure, the post/pre add/remove scripts are missing. Since the systemd rpm macros are broken from my POV I'm keeping the cron scripts. Switching from cron to systemd is something for the user.

PR #156, opened 2015-03-03
- allow to set userdata for snapshots created by helper programs

aschnell:


client/installation-helper.cc
```diff
@@ -254,8 +259,11 @@ main(int argc, char** argv)
     if ((opt = opts.find("description")) != opts.end())
 	description = opt->second;
 
+    if ((opt = opts.find("userdata")) != opts.end())
+	userdata = read_userdata(opt->second);
```
lslezak: Does it make sense to read the user data even if it is not needed? I'd move the reading code into the `step == "1"` condition where it is used...

aschnell: If the option is passed it should be checked for parse errors, so yes.

lslezak:
LGTM

PR #160, opened 2015-04-15
- hackweek 12 improvements

aschnell:


client/snapper.cc
```diff
@@ -95,9 +95,74 @@ struct MyFiles : public Files
...
+
+    void bulk_process(FILE* file, std::function<void(File& file)> callback);
+
 };
 
 
+void
+MyFiles::bulk_process(FILE* file, std::function<void(File& file)> callback)
+{
+    if (file)
```
jreidinger: NP: method is in fact just big if, so maybe it can be splitted to three private methods like
```
if (file)
  bulk_process_file(file, callback)
else if getopts.numArgs() == 0
  bulk_process_noargs(file, callback)
else
  bulk_process_args(file, callback)
```
because otherwise branches do not share code.

jreidinger:
LGTM

PR #163, opened 2015-05-15
Add snapshots creation to installation helpers

imobach:
Add a new step to installation helpers to create an snapshot without D-Bus (to be used during installation). It's will be used by yast-yast2 (see yast/yast-yast2#352 for more information).

### DIFF
--- a/client/installation-helper.cc
+++ b/client/installation-helper.cc
@@ -196,27 +196,34 @@ step4()
 }
 
 void
-step5(const string& description)
+step5(const string& root_prefix, const string& snapshot_type, unsigned int pre_num, const string& description, const map<string, string>& userdata)
 {
     // fate #317973
-    // step runs in chroot
 
     // preconditions (maybe incomplete):
-    // snapper rpms installed in chroot
+    // snapper rpms installed
 
     unsigned int num;
-
+    Snapshots::iterator snapshot;
     SCD scd;
-    scd.read_only = false;
-    scd.description = description;
 
-    Snapper snapper("root", "/");
+    scd.read_only = true;
+    scd.description = description;
+    scd.userdata = userdata;
+
+    Snapper snapper("root", root_prefix);
 
     try
     {
-        Snapshots::iterator snapshot = snapper.createSingleSnapshot(scd);
-        num = snapshot->getNum();
-        snapper.getFilesystem()->setDefault(num);
+        if (snapshot_type == "single") {
+            snapshot = snapper.createSingleSnapshot(scd);
+        } else if (snapshot_type == "pre") {
+            snapshot = snapper.createPreSnapshot(scd);
+        } else if (snapshot_type == "post") {
+            Snapshots snapshots = snapper.getSnapshots();
+            Snapshots::iterator pre = snapshots.find(pre_num);
+            snapshot = snapper.createPostSnapshot(pre, scd);
+        }
     }
     catch (const runtime_error& e)
     {
@@ -224,7 +231,7 @@ step5(const string& description)
         exit(EXIT_FAILURE);
     }
 
-    cout << num << endl;
+    cout << snapshot->getNum() << endl;
     exit(EXIT_SUCCESS);
 }
 
@@ -258,6 +265,8 @@ main(int argc, char** argv)
 	{ "root-prefix",		required_argument,	0,	0 },
 	{ "default-subvolume-name",	required_argument,	0,	0 },
 	{ "description",		required_argument,	0,	0 },
+	{ "snapshot-type",		required_argument,	0,	0 },
+	{ "pre-num",     		required_argument,	0,	0 },
 	{ "userdata",			required_argument,	0,	'u' },
 	{ 0, 0, 0, 0 }
     };
@@ -267,6 +276,8 @@ main(int argc, char** argv)
     string root_prefix = "/";
     string default_subvolume_name;
     string description;
+    string snapshot_type = "single";
+    unsigned int pre_num;
     map<string, string> userdata;
 
     GetOpts getopts;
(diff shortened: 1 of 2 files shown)


jreidinger:
Looks good for me

mvidner:
>  It's will be used by yast-yast2 (see yast/yast-yast2#352 for more information).

This is important context so it should be mentioned in code or a commit message, not lost in a PR discussion. (It often happens to me too, to add the best info only to the PR description)

Any bnc or fate reference?

imobach:
@mvidner You're right. I've added the fate reference to the code.

client/installation-helper.cc
```diff
@@ -195,6 +195,39 @@ step4()
...
+    // fate #317973
+    // step runs in chroot
+
+    // preconditions (maybe incomplete):
+    // snapper rpms installed in chroot
+
+    unsigned int num;
+
+    SCD scd;
+    scd.read_only = false;
```
aschnell: read_only must be true in this case.

client/installation-helper.cc
```diff
@@ -195,6 +195,39 @@ step4()
...
+    scd.description = description;
+
```
aschnell: You should also set the userdata.

client/installation-helper.cc
```diff
@@ -195,6 +195,39 @@ step4()
...
+    Snapper snapper("root", "/");
+
+    try
+    {
+        Snapshots::iterator snapshot = snapper.createSingleSnapshot(scd);
+        num = snapshot->getNum();
+        snapper.getFilesystem()->setDefault(num);
```
aschnell: With that you are already making a rollback. Remove it.

client/installation-helper.cc
```diff
@@ -195,6 +195,39 @@ step4()
...
+
+    SCD scd;
+    scd.read_only = false;
+    scd.description = description;
+
+    Snapper snapper("root", "/");
+
+    try
+    {
+        Snapshots::iterator snapshot = snapper.createSingleSnapshot(scd);
```
aschnell: "single" snapshot is OK for installation but for update you will need a pre and post snapshot.

client/installation-helper.cc
```diff
@@ -195,6 +195,39 @@ step4()
...
+    // preconditions (maybe incomplete):
+    // snapper rpms installed in chroot
+
+    unsigned int num;
+
+    SCD scd;
+    scd.read_only = false;
+    scd.description = description;
+
+    Snapper snapper("root", "/");
```
aschnell: The root_prefix must be flexible so that the tool can also be used for the update case where the "pre" snapshot is done with the binary from the installation medium.

client/installation-helper.cc
```diff
@@ -195,6 +195,39 @@ step4()
     cout << "done" << endl;
 }
 
+void
+step5(const string& description)
+{
+    // fate #317973
+    // step runs in chroot
```
aschnell: See below.

aschnell:
I have the impression that the code is not tested.

client/installation-helper.cc
imobach: Ok, I've misunderstood that.

imobach:
@aschnell I've tried that code in a VM and it created the snapshot. Anyway, I see that I've got a lot of things to fix. I'll ping you when I'm done.

Thanks!

aschnell:
OK, tested in isolation but not in the scope of the feature.

imobach:
@aschnell Yes, in isolation. I've not finished writing the integration tests.

imobach:
I've submitted some fixes and I've also added support for creating pre/post/single snapshots. Could you take a look at these changes, @aschnell ?

Anyway, I'll start now to write proper integration tests.

client/installation-helper.cc
```diff
@@ -195,6 +195,46 @@ step4()
...
 
+void
+step5(const string& root_prefix, const string& snapshot_type, unsigned int pre_num, const string& description, const map<string, string>& userdata)
+{
+    // fate #317973
+
+    // preconditions (maybe incomplete):
+    // snapper rpms installed
+
+    unsigned int num;
```
aschnell: unused variable

client/installation-helper.cc
```diff
@@ -234,6 +276,8 @@ main(int argc, char** argv)
     string root_prefix = "/";
     string default_subvolume_name;
     string description;
+    string snapshot_type = "single";
+    unsigned int pre_num;
```
aschnell: Please initialise to 0 for consistent behaviour if the option "pre-num" is not provided.

client/installation-helper.cc
```diff
@@ -195,6 +195,46 @@ step4()
     cout << "done" << endl;
 }
 
+void
+step5(const string& root_prefix, const string& snapshot_type, unsigned int pre_num, const string& description, const map<string, string>& userdata)
```
aschnell: Please keep lines below 147 chars.

client/installation-helper.cc
```diff
@@ -195,6 +195,46 @@ step4()
...
+        }
+    }
+    catch (const runtime_error& e)
+    {
+        y2err("create snapshot failed, " << e.what());
+        exit(EXIT_FAILURE);
+    }
+
+    cout << snapshot->getNum() << endl;
+    exit(EXIT_SUCCESS);
```
aschnell: The other stepX functions return instead of calling exit. Using exit in C++ will not call all destructors, http://en.cppreference.com/w/cpp/utility/program/exit.

client/installation-helper.cc
```diff
@@ -195,6 +195,46 @@ step4()
     cout << "done" << endl;
 }
 
+int
```
aschnell: Since EXIT_SUCCESS is 0 which evaluates to false and EXIT_FAILURE is 1 which evaluates to true it's better to use a bool as return type and "return step5(...) ? EXIT_SUCCESS : EXIT_FAILURE" in main.

aschnell:
Changelog entry is missing.

aschnell:
LGTM

PR #177, opened 2015-07-15
Handle fpathconf() errors.

jmaggard10:
In one place snapper calls fpathconf(.., _PC_NAME_MAX) to determine the size of
a buffer used to hold readdir_r() results.  fpathconf() may return -1 on error,
but this fpathconf() call's return value is unchecked.  This can result in
allocating a buffer that is too small for readdir_r()'s results, resulting in
out-of-bounds memory access.

Fix it by falling back to using NAME_MAX if fpathconf(.., _PC_NAME_MAX) fails.

Signed-off-by: Justin Maggard <jmaggard@netgear.com>

### DIFF
--- a/snapper/FileUtils.cc
+++ b/snapper/FileUtils.cc
@@ -207,7 +207,10 @@ namespace snapper
 
 	vector<string> ret;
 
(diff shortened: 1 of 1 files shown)


aschnell:
Thanks for the patch.

PR #183, opened 2015-07-29
Transfer Files list via negotiated pipe

oniko:
Hi Arvin,

this is patch proposal for the bug discussed earlier involving too large messages passed over dbus.

The fix includes new dbus method on snapper daemon side:

"GetFilesBySocket"
client asks for Files in similar way as before but it will receive open socket fd if the preparation is successful  instead of whole Files container.

The moment client receives the socket, it can start reading it for the data from the server.

From snapperd perspective:
After sending the dbus reply including the socket fd, snapper daemon will initiate async transfer of Files via the socket. It's a new background thread (per Client instance). The transfer will continue as long as client socket-end is open, or until client disconnects from dbus. During the background transfer the big_mutext is unlocked. I had to add RefCounter for Comparison class so that client request can't delete the comparion before any transfer is finished.

key features:
a) the daemon is less resource hungry now. Files are sent in sets of at most 64K files via buffer
b) the transfer is generally faster thanks to less memory copying between system dbus daemon and clients

the biggest disadvantages so far:
a) serialization and deserialization is performed via boost implemetation. Every third party client would have to reimplement deserialization of a stream sent via negotiated socket.

(apologies in advance for my c++ coding style)

Anyway, please let me know what you think about it in general


oniko:
One more thing...I tried to fix the snapperd abort when client asks for too large data via GetFiles method. So far without success. I tracked it down to failure inside dbus_connection_pop_message() after the reply including too large data is sent from another Client thread.

I'm not sure how to do it properly. The reply is actually sent so I can't detect the failure in request handler.

Perhaps the connection itself is invalidated after trying to send too large data?

oniko:
(socket pair may perform slightly better but it's better not to have open write fd back to snapperd )

server/Client.cc
```diff
@@ -1291,6 +1318,52 @@ Client::get_files(DBus::Connection& conn, DBus::Message& msg)
...
+Client::get_files_pipe(DBus::Connection& conn, DBus::Message& msg)
+{
+    string config_name;
+    string path; // relative to volume root
+    dbus_uint32_t num1, num2;
+
+    DBus::Hihi hihi(msg);
+    hihi >> config_name >> path >> num1 >> num2;
+
+    y2deb("GetFilesByPipe config_name:" << config_name << " subpath: " << path << " num1:" << num1 << " num2:" << num2);
```
aschnell: subpath -> path

server/Client.cc
```diff
@@ -1291,6 +1318,52 @@ Client::get_files(DBus::Connection& conn, DBus::Message& msg)
...
+
+void
+Client::get_files_pipe(DBus::Connection& conn, DBus::Message& msg)
+{
+    string config_name;
+    string path; // relative to volume root
+    dbus_uint32_t num1, num2;
+
+    DBus::Hihi hihi(msg);
+    hihi >> config_name >> path >> num1 >> num2;
```
aschnell: How is the path parameter to be used? Shouldn't there also be a depth parameter?

E.g. then the client could request the files of the root with path=/ and depth=1. Later the client could request the files one level below.

client/commands.cc
```diff
@@ -362,17 +365,34 @@ list<XFile>
...
+	void operator()(XFile* f) { files.push_back(*f); delete f; }
+	list<XFile>& files;
+    };
+
+    Ffunct ff(files);
+
+    while (rs.incoming())
+    {
+	vector<XFile *> tmp = rs.receive();
+	std::for_each(tmp.begin(), tmp.end(), ff);
```
aschnell: Here a lambda function would make the code shorter, e.g.:
```c++
std::for_each(tmp.begin(), tmp.end(), [&files](XFile* f) {
    files.push_back(*f);
    delete f;
});
```

aschnell:
The real disadvantage I see in using the boost serialization is that clients not written in C++ must even implement that part. Couldn't a simpler or more standardised method be used?

oniko:
Hi Arvin,

my apologies for a late reply and thank you for all notes above. Yes, I was aware it would be difficult for 3rd party software to use it, but I had to fix it quick at least for snapper clu.

If you agree the pipe IPC is worth the effort (I mean in general for upstream snapper) I can implement and document the protocol for transferring the files list over the pipe (dropping the boost serialization lib). In the end it's simple text format given that most of data consist only of file paths accompanied by status numbers.

...Or I can make the snapper transfer some well known format: let's say JSON format (for mostly text data it seems to be suitable).

oniko:
(deleted the branch accidentaly, restoring for reference)

PR #188, opened 2015-08-26
Verbose delete, try 2

chrekh:
This is a serie patches to produce verbose output if using arg -v to
snapper cleanup, or snapper delete.


chrekh:
I messed up. This contains bad indentation.

PR #189, opened 2015-08-26
Verbose delete, try 3

chrekh:
This is a serie patches to produce verbose output if using arg -v to
snapper cleanup, or snapper delete.

### DIFF
--- a/client/commands.cc
+++ b/client/commands.cc
@@ -21,6 +21,8 @@
 
 
 #include "commands.h"
+#include "utils/text.h"
+#include "misc.h"
 
 
 #define SERVICE "org.opensuse.Snapper"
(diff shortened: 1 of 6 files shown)


client/commands.cc
```diff
@@ -261,13 +263,23 @@ command_create_post_xsnapshot(DBus::Connection& conn, const string& config_name,
...
-			  list<unsigned int> nums)
+			  list<unsigned int> nums, bool verbose)
 {
     DBus::MessageMethodCall call(SERVICE, OBJECT, INTERFACE, "DeleteSnapshots");
 
     DBus::Hoho hoho(call);
     hoho << config_name << nums;
 
+    if (verbose) {
+	cout << _("Deleting snapshots from ") << config_name << ": ";
```
aschnell: Such sentence fragments can cause trouble for translators. It's better to use ```cout << sformat(_("Deleting snapshots from %s:"), config_name.c_str()) << endl;``` or even the plural form ```cout << sformat(_("Deleting snapshot from %s:", "Deleting snapshots from %s:", nums.size()),	config_name.c_str()) << endl;```.

aschnell:
Thanks, I'll merge it now and improve the translation string as mentioned above.

client/commands.cc
chrekh: Yes, that is much better. Thanks.

PR #198, opened 2015-09-22
- added program to create a subvolume

aschnell:


client/mksubvolume.cc
```diff
@@ -0,0 +1,420 @@
...
+
+bool set_nocow = false;
+bool verbose = false;
+
+
+void
+do_tmp_mount(const libmnt_fs* fs, const char* tmp_mountpoint, const string& subvol_option)
+{
+    if (verbose)
+	cout << "do-tmp-mount" << endl;
```
shundhammer: How about using __FUNCTION__ for this, preferably in a macro?
e.g.

#define TRACE() if (verbose) cout << __FUNCTION__ << endl

client/mksubvolume.cc
```diff
@@ -0,0 +1,420 @@
...
+
+    libmnt_fs* x = mnt_copy_fs(NULL, fs);
```
shundhammer: Poor variable naming. I'd prefer something like 'fs', 'filesystem', 'filesys'. I had to look it up several times what again this 'x' is.

client/mksubvolume.cc
```diff
@@ -0,0 +1,420 @@
...
+    if (!x)
+	throw runtime_error("mnt_copy_fs failed");
+
+    struct libmnt_context* cxt = mnt_new_context();
+    mnt_context_set_fs(cxt, x);
+    mnt_context_set_target(cxt, tmp_mountpoint);
+    if (!subvol_option.empty())
+	mnt_context_set_options(cxt, ("subvol=" + subvol_option).c_str());
+    else
+	mnt_context_set_options(cxt, "subvolid=5");
```
shundhammer: really literal 5 here?

client/mksubvolume.cc
```diff
@@ -0,0 +1,420 @@
...
+void
+do_create_subvolume(const string& tmp_mountpoint, const string& subvolume_name)
+{
+    if (verbose)
+	cout << "do-create-subvolume" << endl;
+
+    string parent = tmp_mountpoint + "/" + dirname(subvolume_name);
+
+    if (mkdir(parent.c_str(), 0777) != 0 && errno != EEXIST)
+	throw runtime_error_with_errno("mkdir failed", errno);
```
shundhammer: not ST_THROW() since this was introduced here in other places?

client/mksubvolume.cc
```diff
@@ -0,0 +1,420 @@
...
+
+    int fd = open(parent.c_str(), O_RDONLY);
+    if (fd < 0)
+	throw runtime_error_with_errno("open failed", errno);
```
shundhammer: again: not ST_THROW() ?

client/mksubvolume.cc
```diff
@@ -0,0 +1,420 @@
...
+
+void
+do_tmp_umount(const libmnt_fs* fs, const char* tmp_mountpoint)
+{
+    if (verbose)
+	cout << "do-tmp-umount" << endl;
+
+    system("/sbin/udevadm settle --timeout 20");
+
+    libmnt_fs* x = mnt_copy_fs(NULL, fs);
```
shundhammer: same complaint as above: better use 'fs' etc. instead of 'x'

client/mksubvolume.cc
aschnell: Yes, that is the initially top-level subvolume with btrfs.

client/mksubvolume.cc
aschnell: Sure, but on the other hand I don't like defines.

client/mksubvolume.cc
aschnell: The program does not setup logging functions so ST_THROW does not bring benefits.

client/mksubvolume.cc
aschnell: Unfortunately the parameter is already called fs. Having e.g. fs and filesystem is not better.

shundhammer:
LGTM

PR #201, opened 2015-09-22
- updated translations

aschnell:


shundhammer:
Although I don't speak ~98% of the languages used there, LGTM.

PR #203, opened 2015-10-14
- removed broken translations file

aschnell:
No changelog needed. The broken file was also pushed without a changelog. Also see https://bugzilla.suse.com/show_bug.cgi?id=950313.


imobach:
LGTM

PR #211, opened 2015-11-24
From Git to OBS with fewer dependencies.

mvidner:
https://trello.com/c/Rg7bI6rP/367-5-add-libstorage-snapper-to-jenkins
The basic idea is to automate the Git -> OBS workflow, including the dependencies needed to make the tarball.

To be used in Jenkins as
```sh

cd $WORKSPACE
rake osc:build  # or osc:commit or osc:sr
```
See https://ci.suse.de/job/snapper/28/console

Yes, Rake (and Ruby) is foreign to C++ development. I think it is justified to introduce it for the specific task of submitting to the openSUSE Build Service, something a C++ contributor does not need but our team does.

A bash-only part that does not rely on Ruby has been split out into https://github.com/openSUSE/obs-package-from-git (see its Readme there).


mvidner:
BTW the shell code is a simplified version of https://github.com/mvidner/maintenance-made-easy

git-to-obs-package
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
jreidinger: hmm, having script that can be potentionally shared between more projects bundled in each git repo can cause maintenance nightmare if we use it more often

Makefile.ci
```diff
@@ -0,0 +1,19 @@
...
+# make -f Makefile.ci osc:ci
+# make -f Makefile.ci osc:build
+# make -f Makefile.ci package
+
+.PHONY: package
+
+package:
+	./git-to-obs-package \
+		-P YaST:Head -p snapper \
+		-c 'make -f Makefile.repo && make package'
```
jreidinger: where is target project? we should also submit it in ci to target repo. Also to support internal BS we need to pass api URL ( so all features we found required by packaging tasks rake itself )

git-to-obs-package
kobliha: +1

mvidner: Fair point, as we want to use it for https://github.com/openSUSE/libstorage in this same PBI. So I have split it off into https://github.com/openSUSE/obs-package-from-git

mvidner:
Now the main work is done in a script that is wgot from a dedicated repo: https://github.com/openSUSE/obs-package-from-git

As for other targets than `package` and `obs_build`, like `obs_ci`, `obs_sr`: I did not implement them because it felt stupid to duplicate the work done by https://github.com/openSUSE/packaging_tasks/blob/master/lib/tasks/osc.rake . If we do want automatic submission then I think it is appropriate to bring `rake` in . We can label it an "upstream packaging tool" and thus avoid the problem of using a "Ruby tool" for "C++ development".

mvidner:
@aschnell what do you think about this?

Makefile.ci
```diff
@@ -0,0 +1,17 @@
...
+	./$^ \
+		-P YaST:Head -p snapper \
+		-o .obsdir \
+		-c 'make -f Makefile.repo && make package'
+
+# A script that is shared with other projects and should be always refreshed
+URL=https://raw.githubusercontent.com/openSUSE/obs-package-from-git/master/obs-package-from-git
+.PHONY: always_out_of_date
+obs-package-from-git: always_out_of_date
+	wget -O $@ $(URL) && chmod +x $@
```
aschnell: Does the CI on jenkins have access to network? Does that allow people to run evil code accessing the intranet by making a pull request?

aschnell:
Is design document available?

Rakefile
```diff
@@ -0,0 +1,26 @@
...
+require "packaging/configuration"
+# skip 'tarball' task, it's redefined here
+Packaging::Tasks.load_tasks(:exclude => ["tarball.rake"])
+
+Packaging.configuration do |conf|
+  conf.obs_project    = "YaST:Head"
+  conf.obs_sr_project = "openSUSE:Factory"
+  conf.package_dir    = ".obsdir"
+  conf.skip_license_check << /.*/
+end
```
aschnell: Where is this data use?

Makefile.ci
jreidinger: yes, it have. It is needed e.g. for git updates.

jreidinger: and second question. No, just by making pull request it do not create package on jenkins, unless somehow switched.

Rakefile
jreidinger: it is data, that specify where to send package from master. That BTW mean, that it do not automatic send to SP2 as it is implemented now.

Makefile.ci
aschnell: But if we want to run checks also for pull requests like travis ci does now we have a severe security issue, right?

jreidinger: no as we wget always script from master. But of course if you modify this Makefile.ci here, you have problem. Exactly same problem as now if someone create malicious stuff in configure.ac. So it is more problem of running jenkins for such kind of pull request then this pr itself.

aschnell: Yes, of course a pull request can modify Makefile.ci.

So for me the setup is not OK.

jreidinger: but it can even now modify e.g. Makefile.repo, so what is difference?

I think main issue is that if we check on jenkins all pull requests, then we cannot use anything from that git repo, which is not true for any current project, so we have to address it when we decide to add such functionality on jenkins.

Rakefile
mvidner: This config ([API docs here](http://www.rubydoc.info/github/openSUSE/packaging_tasks/Packaging/Configuration)) is used by `rake osc:commit` and `rake osc:sr`. Actually these tasks are not used on ci.suse.de now (only `rake osc:build` is), as I did not want automatic submission before we have agreed on these PRs.

mvidner: Thanks for pointing out SP2, I should unify this package to point there too. (For YaST it was done by https://github.com/yast/yast-rake/pull/26 )

Makefile.ci
mvidner: Thanks for pointing out untrusted code, Arvin. I think we will need to ask the OBS people what the best practices are.
But it will only be a hard limit when we want to start running CI for all pull requests. For the time being, where only merged and reviewed repos are used, I do not see it as an obstacle for merging this.

mvidner: > > Does the CI on jenkins have access to network? 

> needed e.g. for git updates.

I think that could be avoided by having Jenkins fetch the code, then run the job in a container/whatever without network. Then the wget would not indeed work.

mvidner:
> Is design document available?

Um, no, but I'll write it if you tell me what you want to see there. Is https://wiki.microfocus.net/index.php?title=YaST/jenkins a good starting point? What questions should I answer?

Rakefile
jreidinger: ci.opensuse.org do not by default submit to SP2 but to factory. If you are interested, ci.s.d send to SP2

Makefile.ci
jreidinger: no as job also need network for e.g. build where it download the latest packages.

aschnell: Downloading the packages needs network but the build process itself does not. The obs apparently solved that problem.

aschnell:
A design document should explain the overall workflow and what the tasks of Jenkins and of the workers are. I also would like to see why the Packaging.configuration should be saved in the snapper repo and not in Jenkins or some meta data repository (I was surprised when the SP1 branch was made and some data in every YaST repo had to be updated).

jreidinger:
how jenkins can know which branch is for which data? you can only hardcode it there and it destroy idea of pull requests. Also why we create maintenance branches? I see that reason is to maintain version for given product, so it is nature for me to have there note for which product we maintain it ( one liner where it is specified which product is target one ).

aschnell:
Currently Jenkins does know what repositories it has to look at and the repositories define what branch is used for which product. So the information is split while it could also be defined in one meta data repository. Why that should destroys the idea of pull requests is not clear to me.

mvidner:
Arvin said: 
> I was surprised when the SP1 branch was made and some data in every YaST repo had to be updated

Yes, but that will not be necessary since the introduction of the YAST_SUBMIT environment variable.

mvidner:
An analogous PR for libstorage: https://github.com/openSUSE/libstorage/pull/154

imobach:
LGTM

PR #213, opened 2015-12-17
Allow rollback for root file system only 

lslezak:
This is a fix for [bug#954445](https://bugzilla.suse.com/show_bug.cgi?id=954445).

#### Details

The check is similar to the existing `btrfs` check. The fix includes a small refactoring - the snapper config is received only once via DBus and used in both old `getFilesystem()` and the new `getSubVolume()` functions.

#### Test

Tested in SLE12-SP1, prepare the system first:

```
# snapper create-config /
# snapper -c opt create-config /opt
# snapper -c opt create
```

Originally the snapper allowed to rollback the snapshot (resulting in an unbootable system, see the bug). With the fix an error is printed and the rollback is aborted:

```
# snapper -c opt rollback 1
Command 'rollback' cannot be used on a non-root subvolume /opt.
```


aschnell:
LGTM

PR #220, opened 2016-01-18
Support platforms using musl libc

lorenzph:
While packaging snapper for [Alpine Linux](http://alpinelinux.org/) which uses [musl](http://www.musl-libc.org/) as its libc implementation I encountered some portability issues. This patch series enables successful deployment of snapper on targets using musl libc.


stintel:
@lorenzph thanks for your effort
Can someone please review and/or merge this? Thanks!

aschnell:
As far as I see most of the problems this PR addresses have been fixed now. If there are still problems please make new PRs where each PR deals with one problem.

PR #221, opened 2016-01-20
Fix dbus interface methods types.

p2004a:
The XML in Introspect method doesn't contain correct information about a few methods.

- ListConfigs, ListSnapshots, ListSnapshotsAtTime, GetFiles don't return variadic but raw value (not packed inside variadic).
- MountSnapshot, GetMountPoint didn't contain information about return type.
- CreateComparison declared that it returns num of files but implementation didn't return anything.

I tried to add a unit/real test that will validate the interface types but I couldn't mock objects that Client class is using and I can't run snapperd without interfring with snapper currently installed in the system.

I was testing in sandboxed installation using this scripts: https://gist.github.com/p2004a/9762b22c3edeb0f5d18a (pydbus library validates the return types etc.)

### DIFF
--- a/server/Client.cc
+++ b/server/Client.cc
@@ -203,7 +203,7 @@ Client::introspect(DBus::Connection& conn, DBus::Message& msg)
 	"    </signal>\n"
 
 	"    <method name='ListConfigs'>\n"
-	"      <arg name='configs' type='v' direction='out'/>\n"
+	"      <arg name='configs' type='a(ssa{ss})' direction='out'/>\n"
 	"    </method>\n"
 
 	"    <method name='GetConfig'>\n"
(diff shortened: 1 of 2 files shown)


server/Client.cc
```diff
@@ -314,6 +314,7 @@ Client::introspect(DBus::Connection& conn, DBus::Message& msg)
 	"      <arg name='config-name' type='s' direction='in'/>\n"
 	"      <arg name='number' type='u' direction='in'/>\n"
 	"      <arg name='user-request' type='b' direction='in'/>\n"
+	"      <arg name='path' type='s' direction='out'/>\n"
```
aschnell: Please also fix in dbus-protocol.txt.

server/Client.cc
```diff
@@ -325,6 +326,7 @@ Client::introspect(DBus::Connection& conn, DBus::Message& msg)
 	"    <method name='GetMountPoint'>\n"
 	"      <arg name='config-name' type='s' direction='in'/>\n"
 	"      <arg name='number' type='u' direction='in'/>\n"
+	"      <arg name='path' type='s' direction='out'/>\n"
```
aschnell: Please also fix in dbus-protocol.txt.

server/Client.cc
```diff
@@ -1225,6 +1227,10 @@ Client::create_comparison(DBus::Connection& conn, DBus::Message& msg)
 
     DBus::MessageMethodReturn reply(msg);
 
+    DBus::Hoho hoho(reply);
+    dbus_uint32_t num_files = comparison->getFiles().size();
+    hoho << num_files;
+
```
aschnell: Please also fix in dbus-protocol.txt.

aschnell:
Thanks for the patch. Please also fix the mistakes in doc/dbus-protocol.txt.

p2004a:
Done.

aschnell:
Thanks.

Issue #176, opened 2015-07-10
 Checking status with snapper receiving Failure (org.freedesktop.DBus.Error.NoReply)


oniko:
Hi Arvin,

I have some bad news for Friday. The bug I reference is reproducible also with upstream snapper (0.2.6) and it's not related to lvm only. I hope I'm getting close to the solution, will let you know about details when I have some. I fear it's related to DBus interface as being implemented in snapper but again, will provide you with details when I have some.

bug: http://bugzilla.redhat.com/show_bug.cgi?id=1229353

So only (sad) heads up for you:/

Regards
O.

aschnell:
You are hitting the DBus message size limit. So the problem has an easy and a hard fix.

Easy: Increase size limit in /etc/dbus-1/system.conf. E.g. add:
```
  <limit name="max_incoming_bytes">100000000</limit>
  <limit name="max_outgoing_bytes">100000000</limit>
  <limit name="max_message_size">100000000</limit>
```

Hard: Implement other technique to pass the file list from server to client.

Maybe this gets better with kdbus.

oniko:
Thank you for quick reply! Yes, increasing the limit helped workaround the issue, but I'm affraid it's only temporary solution. According to DBus docs, it seems there's a hard limit on message size which is 128 MB. This ~40MB filelist is pretty much too close to that limit. I'm affraid I can avoid the hard fix only for a limited time...

This is immediate idea (and probably wrong): what about introducing interface for splitting large messages among smaller ones:

request: GetFiles
response: "part 1, 8 more will follow"
response: "part 2"
response: "part 3"
...

But yes. This would add much more complexity given I'd have to deal with new error paths...

oniko:
Well. the other way (and probably a lot faster) would be to pass the file-list through socket/pipe negotiated via message bus. Unfortunately we would have to parse the file-list on client-side...

aschnell:
Yes, you have outlined the two basic ideas for a proper fix I see. One interesting modification to the first idea could be to allow clients to query a tree of files instead of a simply list. E.g.
```
GetFilesTree("/")
```
would return
```
[
  { "/etc", status, has_subtree },
  { "/usr", status, has_subtree }
]
```

The client can then continue with GetFilesTree("/etc") and GetFilesTree("/usr").

That could overall reduce the amount of transmitted data since paths are transmitted only once but at the same time increase the number of request.

A client representing the files as a tree to the user anyway could even delay the requests until the user opens a directory.

[...]

PR #223, opened 2016-02-07
Improved handling of dbus disconnection

p2004a:
Related to #176. I was thinking about more clear way of doing it but it looks like every other method is just throwing FatalException when something unexpected happens so I implemented this in the same way.


aschnell:
Thanks for the patch. It did not fit anymore so I reworked and extended it.

p2004a:
Thank you! :slightly_smiling_face:

PR #225, opened 2016-03-15
rule out race while (u)mounting lvm snapshot

oniko:
Hi Arvin, snapper maintainers,

this commit fixes a race in snapperd daemon while trying to (u)mount an lvm snapshot from more than one thread. To reproduce it just run following snapper commands close to each other on lvm backend.

snapper create -t post --pre-num 1 (with background comparison enabled, let's say post snapshot will be 2)
and immediately after it:
snapper status 1..2

The clash is between Background worker and create_comparison thread originating in snapper status cmd.

### DIFF
--- a/snapper/Lvm.cc
+++ b/snapper/Lvm.cc
@@ -251,6 +251,8 @@ namespace snapper
     void
     Lvm::mountSnapshot(unsigned int num) const
     {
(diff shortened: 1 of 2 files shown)


aschnell:
Thanks, looks fine. A patch for the ext4 backend is not needed since that backend is not maintained anymore (it has compile time error since years).

PR #228, opened 2016-03-24
- added wrappers for btrfs quota ioctls

aschnell:


snapper/BtrfsUtils.cc
```diff
@@ -249,18 +251,92 @@ namespace snapper
...
+	    sk->min_type = BTRFS_ROOT_ITEM_KEY;
+	    sk->max_type = BTRFS_ROOT_ITEM_KEY;
+	    sk->min_offset = 0;
+	    sk->max_offset = (u64) -1;
+	    sk->min_transid = 0;
+	    sk->max_transid = (u64) -1;
+	    sk->nr_items = 1;
+
+	    if (ioctl(fd, BTRFS_IOC_TREE_SEARCH, &args) != 0)
+		throw runtime_error_with_errno("ioctl(BTRFS_IOC_TREE_SEARCH) failed", errno);
```
shundhammer: Since this is used so often: maybe use a function or a macro and thus repeat duplication?
Maybe even one that does the ioctl, only receiving fd, the ioctl name, and args?

shundhammer:
LGTM

snapper/BtrfsUtils.cc
aschnell: Yes, the repetition is not nice. But avoiding it requires a macro and those are also not nice.

Issue #219, opened 2016-01-18
DBus::Marshalling has subclasses named Hihi and Hoho


Profpatsch:
I had a big “wat”-moment.

https://github.com/openSUSE/snapper/blob/master/server/Types.h

mvidner:
Oh. Er, um.
Apparently we should rename `Hoho` to `Marshaller` and `Hihi` to `Unmarshaller`.

Profpatsch:
hehe. :)

PR #229, opened 2016-03-30
- work on cleanup using quota groups

aschnell:


client/commands.cc
```diff
@@ -402,6 +403,37 @@ command_get_xfiles(DBus::Connection& conn, const string& config_name, unsigned i
 
 
 void
+command_prepare_quota(DBus::Connection& conn, const string& config_name)
+{
+    DBus::MessageMethodCall call(SERVICE, OBJECT, INTERFACE, "PrepareQuota");
+
+    DBus::Hoho hoho(call);
```
shundhammer: We just talked about this, but just for the record: "Hoho", "hoho", "Hihi", "hihi" are not good names.
Since this is about DBUS communication, I'd suggest something like "request" and "reply".

doc/dbus-protocol.txt
```diff
@@ -53,6 +53,8 @@ The following command require a successful CreateComparison in advance.
 method GetFiles config-name number1 number2 -> list(filename status)
 
 
+Intentionally not documented are PrepareQuota and QueryQuota.
```
shundhammer: Huh?

client/commands.cc
aschnell: See https://github.com/openSUSE/snapper/issues/219.

doc/dbus-protocol.txt
aschnell: Until the implementation (including kernel) has shown to be satisfying I don't want to document the methods so that changing them is possible.

snapper/Btrfs.cc
```diff
@@ -314,7 +318,7 @@ namespace snapper
 #ifdef ENABLE_ROLLBACK
 
     void
-    Btrfs::createSnapshotOfDefault(unsigned int num, bool read_only) const
+    Btrfs::createSnapshotOfDefault(unsigned int num, bool read_only, bool quota) const
```
shundhammer: If those boolean parameters grow any more, maybe OR'ed flags might be a good idea. Not right now with two of them, but if there will be a third, I'd strongly recomend this.

snapper/Btrfs.h
```diff
@@ -59,9 +60,9 @@ namespace snapper
 	virtual SDir openInfosDir() const;
 	virtual SDir openSnapshotDir(unsigned int num) const;
 
-	virtual void createSnapshot(unsigned int num, unsigned int num_parent,
-				    bool read_only) const;
-	virtual void createSnapshotOfDefault(unsigned int num, bool read_only) const;
+	virtual void createSnapshot(unsigned int num, unsigned int num_parent, bool read_only,
```
shundhammer: How about using C++11 'override' for those overwritten virtual methods? Even more so since the interface has now changed, so it would be desirable to make sure they are all overwritten correctly?

aschnell: Yes, but snapper is also build for older distributions. But indeed, ```override``` may be safe to use today.

shundhammer:
Apart from the minor issues mentioned above, LGTM.

But we should really get rid of those Hihi/Hoho thingies in the mid-term future. ;-)

PR #232, opened 2016-04-01
- work on btrfs ioctl wrappers

aschnell:


snapper/BtrfsUtils.cc
```diff
@@ -458,7 +458,13 @@ namespace snapper
 	};
 
 
-	void
+	/*
+	 * Wrapper for ioctl(BTRFS_IOC_TREE_SEARCH). Calls callback of
+	 * tree_search_opts for every found item.  In contrast to the bare
+	 * ioctl the wrapper ensures that the min and max values in
+	 * tree_search_opts are satisfied.
+	 */
```
mvidner: @return number of times the callback was called

mvidner:
The matching FATE is https://fate.suse.com/312751

testsuite-real/test-btrfsutils.cc
```diff
@@ -0,0 +1,106 @@
...
+{
+    int fd = open("/btrfs", O_NOATIME);
+    if (fd < 0)
+    {
+	cerr << "open failed (" << strerror(errno) << ")" << endl;
+	return EXIT_FAILURE;
+    }
+
+
+    if (false)
```
mvidner: Why bother writing a "test" that has almost everything disabled?

snapper/BtrfsUtils.cc
aschnell: No doxygen or alike is involved here. Added return value as English text.

testsuite-real/test-btrfsutils.cc
aschnell: Since these are integration tests and the program must be adapted. I added a comment.

mvidner:
Thanks.
Well, that's unfortunately as much as I am able to review here, as I don't know what quota groups are. Which documentation should I read?

aschnell:
You can read http://sensille.com/qgroups.pdf for the basic of quota group. The ioctl code is based on code in btrfsprogs.

mvidner:
Thanks for the pointer to the docs. It would be useful to include it in a comment in the source.

It will take me a few days to understand the concepts, and until then I am unable to comment on the logic of the code.
Regarding the form, LGTM.

PR #238, opened 2016-04-11
Huha snapperd sigabrt

shundhammer:
https://bugzilla.suse.com/show_bug.cgi?id=940154

Create 'component' string on the heap to make sure it's not destroyed in an undefined order as long as the Snapper object still needs it. Intentionally leaving it behind as a memory leak (it's only some 10-16 bytes anyway).


aschnell:
LGTM

PR #240, opened 2016-04-12
Backport of fix for snapperd crash to SLE-12 SP1

shundhammer:
https://trello.com/c/r21x30Ou/83-5-os-tw-902352-snapperd-crashed-shortly-after-boot
https://bugzilla.suse.com/show_bug.cgi?id=940154

Backporting the fix for bsc#940154 to SLE-12 SP1

This will only be shipped if there is a maintenance update anyway or if this problem hits customers hard.


aschnell:
LGTM

PR #242, opened 2016-05-18
[FR] some API extensions

aleksey-novikov:
For some my local project I need an utility using Snapper API. But for this purpose I needed to expand API.

These changes allow:
- change default snapshot (will be activated on next boot time)
- determine snapshot is default (will be activated on next boot time)
- determine snapshot is active (activated on last boot time)

These changes don't modify behavior of other components.

### DIFF
--- a/snapper/Btrfs.cc
+++ b/snapper/Btrfs.cc
@@ -1451,6 +1451,29 @@ namespace snapper
 	}
     }
 
+
+    bool
+    Btrfs::isActive(unsigned int num) const
+    {
+	bool ret = false;
+
+	try
+	{
+	    if (num == 0)
+		SN_THROW(IllegalSnapshotException());
+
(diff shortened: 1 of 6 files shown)


aleksey-novikov:
For some local project I need an utility using Snapper API. But for this purpose I needed to expand API.
These changes allow:
- change default snapshot (will be activated on next boot time)
- determine snapshot is default (will be activated on next boot time)
- determine snapshot is active (activated on last boot time)

aschnell:
Looks fine. Could you please copy the explanation of default and active from the pull request description to the code. You could also mention that default == active unless the default was changed since bootup.

aleksey-novikov:
@aschnell Well, I'll add a description of these methods. Note, when you manually select another snapshot in bootloader then default != active.

PR #249, opened 2016-07-14
Clean up the .spec file template

pjanouch:
- applied spec-cleaner
 - assumed SUSE everywhere
 - made sure libtool files aren't installed
 - split configuration into another package

### DIFF
--- a/snapper.spec.in
+++ b/snapper.spec.in
@@ -1,7 +1,7 @@
 #
 # spec file for package snapper
 #
-# Copyright (c) 2015 SUSE LINUX GmbH, Nuernberg, Germany.
+# Copyright (c) 2016 SUSE LINUX GmbH, Nuernberg, Germany.
 #
 # All modifications and additions to the file contributed by third parties
 # remain the property of their copyright owners, unless otherwise agreed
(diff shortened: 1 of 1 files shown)


aschnell:
Fails at least with Fedora 23.

pjanouch:
The current version fails as well for both Fedora 23 and 24 on OBS, and I removed the SUSE checks intentionally as advised by @scarabeusiv . RedHat/Fedora maintain their own spec file for snapper.

aschnell:
With some research you could find that SUSE provides packages also for Fedora, see https://build.opensuse.org/project/monitor/filesystems%3Asnapper.

pjanouch:
That is for an outdated version, though.

aschnell:
With minor adjustments it works.

The product owner even wants to automate submission to that project (https://trello.com/c/HnurGpsR/688-3-resolve-the-double-development-project-for-snapper) so any change making the situation worse is not acceptable.

scarabeusiv:
Ok just few points:

1) fedora does not use it and provide their own package from within their repository http://pkgs.fedoraproject.org/cgit/rpms/snapper.git/
2) that trello board is inaccesible so we can't tell you what is in there and if it make sense
3) whom is the product owner so I could talk to him? Also I still think the openSUSE developement and quality of openSUSE packages take priority over providing package to our competing distribution.

aschnell:
Just a few remarks:

1. We also provide packages for other distributions as already mentioned. This was requested (not by fate AFAIR) by Matthias Eckermann years ago to promote snapper.

2. Sorry for that. Please talk to the product owner Jiri Srain.

3. The product owner is Jiri Srain. And finally: Portability is an important issue for software we develop at SUSE, esp. since the software is included in many distributions, and the build service project is also used to check the portability.

So please talk with the product owner before reopening.

Issue #256, opened 2016-08-04
snapperd: abrtd detects error when snapperd finishes running on debug mode


oniko:
Hi Arvin,

originally reported in https://bugzilla.redhat.com/show_bug.cgi?id=1357010 but I can easily reproduce it also with current 0.3.3.

There's clash between lifetime of static mutex in simple_log_do() and global Backgrounds object. It happens only when daemon is shutting down and it was started with -d argument in debug mode. Also you have to create workload with at least single background job so that y2deb("worker interrupted") is hit during Backgrounds destruction to reproduce it.

I'm not too familiar with C++ std, but so far I haven't figured out different bulletproof solution than removing the y2deb() call from the Backgrounds worker. Which is not optimal since we'd like to know that someone or something has interrupted one of snapperd threads. All other solution I tried are undefined by the c++ standard:/

Anyway I'll let you know If I come with saner patch other than the one deleting the debug message...

oniko:
Feel free to use it or discard it, as usual. I'm not sure there is a better solution while taking into consideration global objects in various translation units... I found myself in 'undefined behaviour' territory with every other attempts I made.

Kind regards
O.

aschnell:
Merged #258 to fix issue.

PR #258, opened 2016-08-08
fix objects termination clashes on shutdown

oniko:
This is one of possible fixes for issue #256. My goal is to make it clear in what order objects used solely in snapperd server are torn down.

Of course the minimalistic fix is to remove y2deb() message from Backgrounds worker invoked on snapperd shutdown.

Regards O.

### DIFF
--- a/server/Client.cc
+++ b/server/Client.cc
@@ -37,11 +37,9 @@
 
 boost::shared_mutex big_mutex;
 
-Clients clients;
 
-
-Client::Client(const string& name)
-    : name(name), zombie(false)
+Client::Client(const string& name, const Clients& clients)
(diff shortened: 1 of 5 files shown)


server/Client.cc
```diff
@@ -1726,6 +1724,19 @@ Client::worker()
...
+Clients::Clients(Backgrounds& backgrounds)
+    : bgs(backgrounds)
+{
+}
+
+
+Backgrounds&
+Clients::backgrounds() const
+{
+    return bgs;
```
aschnell: Such simple functions can usually go into the header file.

aschnell:
Thanks for the patch, it looks very fine.

Issue #257, opened 2016-08-08
on cleanup: "Client.cc(dispatch):1684 - caught unknown exception"


saladpanda:
System-info:

```
❯ snapper --version
snapper 0.3.2
flags btrfs,lvm,no-ext4,xattrs,rollback,btrfs-quota

~
❯ uname -a
Linux suse 4.7.0-1-default #1 SMP PREEMPT Mon Jul 25 08:42:47 UTC 2016 (89a2ada) x86_64 x86_64 x86_64 GNU/Linux

~
❯ cat /etc/issue
Welcome to openSUSE 20160805 "Tumbleweed" - Kernel \r (\l).
```

I'm getting the following error:

```
❯ sudo snapper cleanup number   
Failure (error.something).

❯ sudo tail -30 /var/log/snapper.log
2016-08-08 10:56:05 MIL libsnapper(6708) Snapper.cc(loadIgnorePatterns):156 - number of ignore patterns:8
2016-08-08 10:56:05 MIL libsnapper(6708) Snapshot.cc(read):224 - found 9 snapshots
2016-08-08 10:56:05 ERR libsnapper(6708) Client.cc(dispatch):1684 - caught unknown exception
2016-08-08 10:56:35 MIL libsnapper(6708) Snapper.cc(~Snapper):118 - Snapper destructor
2016-08-08 10:57:05 MIL libsnapper(6708) snapperd.cc(main):282 - Exiting
2016-08-08 11:00:58 MIL libsnapper(8747) snapperd.cc(main):270 - Requesting DBus name
2016-08-08 11:00:58 MIL libsnapper(8747) snapperd.cc(main):274 - Loading snapper configs
2016-08-08 11:00:58 MIL libsnapper(8747) Snapper.cc(getConfigs):251 - Snapper get-configs
2016-08-08 11:00:58 MIL libsnapper(8747) Snapper.cc(getConfigs):252 - libsnapper version 0.3.2
2016-08-08 11:00:58 MIL libsnapper(8747) AsciiFile.cc(reload):114 - loading file /etc/sysconfig/snapper
2016-08-08 11:00:58 MIL libsnapper(8747) AsciiFile.cc(getValue):235 - key:SNAPPER_CONFIGS value:root
2016-08-08 11:00:58 MIL libsnapper(8747) AsciiFile.cc(reload):114 - loading file /etc/snapper/configs/root
[...]

PR #259, opened 2016-08-08
- better messages for some btrfs quota errors (gh#openSUSE/snapper#257)

aschnell:
To address bad error message in #257.


wfeldt:
lgtm

PR #261, opened 2016-08-12
replace getmntent with getmntent_r

oniko:
Hi Arvin,

this is follow up to reported race while mounting same snapshot simultaneously (merged pull request #225). Unfortunately my original fix was not complete. It's not wrong but I overlooked the fact that snapper uses getmntent() call which is not thread safe. The use of this function caused that sometimes Lvm::IsSnapshotMounted() method returns 'false' in case while other thread already successfully mounted the snapshot. The lock I implemented has mitigated the issue only. The proper fix should include this commit as well.

The only test so far I have is to run the original reproducer for a long time.

### DIFF
--- a/snapper/AppUtil.cc
+++ b/snapper/AppUtil.cc
@@ -36,6 +36,7 @@
 #include <mntent.h>
 #include <boost/algorithm/string.hpp>
 #include <boost/io/ios_state.hpp>
+#include <boost/scoped_array.hpp>
 
 #include "snapper/Log.h"
 #include "snapper/AppUtil.h"
@@ -170,6 +171,14 @@ namespace snapper
     }
 
 
+    unsigned
+    pagesize()
+    {
+	long r = sysconf(_SC_PAGESIZE);
+	return r < 0 ? 4096 : r;
+    }
+
+
     bool
     getMtabData(const string& mount_point, bool& found, MtabData& mtab_data)
     {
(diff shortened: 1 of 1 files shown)


snapper/AppUtil.cc
```diff
@@ -182,19 +191,23 @@ namespace snapper
 
 	found = false;
 
-	struct mntent* m;
-	while ((m = getmntent(f)))
+	struct mntent m;
+	// each mnt table element is limited to PAGE_SIZE top in Linux
```
aschnell: Mainly for my curiosity: Where is that documented?

snapper/AppUtil.cc
```diff
@@ -182,19 +191,23 @@ namespace snapper
...
+	unsigned buf_size = 4 * pagesize();
```
aschnell: Why multiply with 4?

snapper/AppUtil.cc
oniko: Hello,

well I have to admit it's not (at least I can't find it directly mentioned anywhere). So I'm sorry for the misleading comment (I've mistaken /proc for /sys attributes in my mind). But I think it's good  compromise anyway. The mnt ent elements are as follows:

a) device is path, or UUID (PATH_MAX?)
b) mnt dir is path (PATH_MAX?)
c) fs type ???
d) mnt options ???. well, there're filesystems accepting an option with path as an argument so....PATH_MAX +more?

I've seen various buffer sizes in use with getmntent_r: 2*PATH_MAX (wrong due to above)
single PAGE_SIZE - wrong (wrong due to above). This is in glibc getmntent implementation btw (which calls getmntent_r internally).

But I'm open to suggestions

snapper/AppUtil.cc
oniko: yup

aschnell:
Thanks for the patch.

oniko:
Well I'm late to the party but would you mind me replacing getmntent(_r) with libmount functions? there's libmount dependency in snapper for btrfs anyway and we'd delegate the problem to more appropriate library:)

I'd build it around  mnt_new_table_from_file("/proc/self/mountinfo"). But I'm not sure if there're some btrfs specific issues that would introduce new bug while fixing this one...

oniko:
(provided those calls are thread safe, of course)

aschnell:
In general I would not mind replacing handmade code with libmount functions.

Issue #264, opened 2016-08-21
Feature: Boot Snapshots


NicoHood:
Is it possible that the snapper package provides boot snapshot service + timer by default? Similar to this one:
https://wiki.archlinux.org/index.php/Snapper#Snapshots_on_boot

PR #267, opened 2016-09-01
Added boot service

NicoHood:
Fixes https://github.com/openSUSE/snapper/issues/264

The number cleanup algorithm is used for the boot configuration which makes most sense to me. I've also recently added this to the arch wiki. `-c root` clarifys the root setting, however it could also be left out.


NicoHood:
Done. Thanks, I've missed that.

aschnell:
Thanks.

PR #275, opened 2016-10-01
It'd be better to install timers.target.wants

henrich:
How about setting those timers as [Install] WantedBy=timers.target?

### DIFF
--- a/data/boot.service
+++ b/data/boot.service
@@ -2,5 +2,5 @@
 Description=Take snapper snapshot of root on boot
 
 [Service]
-Type=simple
+Type=oneshot
 ExecStart=/usr/bin/snapper --config root create --cleanup-algorithm number --description "boot"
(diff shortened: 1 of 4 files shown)


NicoHood:
I've no idea if the PR is correct, but if so you should also edit the new boot timer:
https://github.com/openSUSE/snapper/blob/master/data/boot.timer

henrich:
On Mon, 03 Oct 2016 05:25:07 -0700
Nico <notifications@github.com> wrote:
> I've no idea if the PR is correct, but if so you should also edit the new boot timer:
> https://github.com/openSUSE/snapper/blob/master/data/boot.timer

 Probably yes, I thought if my proposal is okay, then will check
 boot timer. But maybe it is okay to merge both changes.

aschnell:
The change looks OK to me after reading https://www.freedesktop.org/software/systemd/man/systemd.special.html. Please also change boot.timer as NicoHood already mentioned.

aschnell:
Thanks for the patch.

PR #278, opened 2016-10-07
Add bash completion

henrich:
Taken from my Debian package. Not perfect, but not bad for stating point.

### DIFF
--- a/scripts/Makefile.am
+++ b/scripts/Makefile.am
@@ -19,6 +19,7 @@ EXTRA_DIST = snapper-hourly snapper-daily zypp-plugin.py $(pam_snapper_SCRIPTS)
 install-data-local:
 	install -D snapper-hourly $(DESTDIR)/etc/cron.hourly/suse.de-snapper
 	install -D snapper-daily $(DESTDIR)/etc/cron.daily/suse.de-snapper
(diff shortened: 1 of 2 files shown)


scripts/Makefile.am
```diff
@@ -19,6 +19,7 @@ EXTRA_DIST = snapper-hourly snapper-daily zypp-plugin.py $(pam_snapper_SCRIPTS)
 install-data-local:
 	install -D snapper-hourly $(DESTDIR)/etc/cron.hourly/suse.de-snapper
 	install -D snapper-daily $(DESTDIR)/etc/cron.daily/suse.de-snapper
+	install -D bash-completion.sh $(DESTDIR)/usr/share/bash-completion/completions/snapper
```
NicoHood: You also need to add this file to EXTRA_DIST in line 17 if i am not completely wrong. Thats what I also was adviced in my boot timer PR.

NicoHood:
As he said, no perfect but a good starting point. I've not tested the makefile, but the completion itself which seems to work. I've noticed the -c option does not privde the config, nor does it continue to suggest new commands afterwards.

It would be nice to add this first start of bash completion.

Issue #286, opened 2016-11-01
setup-quota fail


NicoHood:
When I use snapper along with quotas It fails for the 2nd config that I try to enable quotas for. I've been testing this on the raspberry pi 3 with kernel 4.4 on archlinux for arm. I've installed my system like this:
https://github.com/NicoHood/NicoHood.github.io/wiki/Raspberry-Pi-Encrypted-Btrfs-Root

After that fresh install I've enabled snapper as the following:
```bash
# Setup configurations
sudo pacman -S snapper snap-pac
sudo umount /.snapshots
sudo rm -r /.snapshots
sudo snapper -c root create-config /
sudo snapper -c home create-config /home
sudo mount -a
sudo chmod 750 /.snapshots
sudo chown :wheel /.snapshots
sudo chown :users /home/.snapshots

# Start snapper timers
sudo systemctl enable snapper-timeline.timer
sudo systemctl enable snapper-cleanup.timer
sudo systemctl start snapper-timeline.timer
sudo systemctl start snapper-cleanup.timer

# Allow users to use the snapper -c root list command
sudo snapper set-config ALLOW_GROUPS=wheel
sudo snapper -c home set-config ALLOW_GROUPS=wheel

# Enable snapper quota
snapper setup-quota
snapper -c home setup-quota
```

The 2nd command fails like this:
```
[alarm@alarmpi ~]$ snapper -c home setup-quota
Failure (error.something).
[alarm@alarmpi ~]$ snapper -c home setup-quota
Failure (org.freedesktop.DBus.Error.NoReply).
[alarm@alarmpi ~]$ snapper -c home setup-quota
Failure (org.freedesktop.DBus.Error.NoReply).
```

I've also had this in my other installation where my btrfs then had unfixable errors. I have not checked the btrfs yet, but there must be something wrong with it.

Do you have any ideas why? I am using snapper 0.3.3-3.

NicoHood:
The output of the qgroups looks like this:

```
[alarm@alarmpi ~]$ sudo btrfs qgroup show /
WARNING: qgroup data inconsistent, rescan recommended
qgroupid         rfer         excl 
--------         ----         ---- 
0/5          16.00KiB     16.00KiB 
0/257         1.16GiB     16.48MiB 
0/258        16.00KiB     16.00KiB 
0/259         1.00MiB      1.00MiB 
0/260        16.00KiB     16.00KiB 
0/261       247.80MiB    247.80MiB 
[...]

PR #288, opened 2016-11-02
- fixed use of getpwuid_r, getpwnam_r and getgrnam_r

aschnell:
Reported in https://lists.opensuse.org/yast-devel/2016-11/msg00002.html.

### DIFF
--- a/package/snapper.changes
+++ b/package/snapper.changes
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Wed Nov 02 22:17:16 CET 2016 - aschnell@suse.com
+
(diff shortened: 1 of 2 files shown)


shundhammer:
LGTM

NicoHood:
Does this also affect those two commands?
```
sudo snapper set-config ALLOW_GROUPS=wheel
sudo snapper -c home set-config ALLOW_GROUPS=wheel
```

I am wondering if this has something to do with #286

aschnell:
It could affect those commands (in general everything dealing with users or groups) but only when the user and group setup causes the structures provided by those libc calls to be big. This could be due to e.g. very long names or groups with many users.

PR #293, opened 2016-11-16
- added prototype of proxy classes to work with and without dbus

aschnell:
For https://trello.com/c/lOl4PmBF/475-5-fate-321049-prototype-proxy-classes-for-d-bus-no-d-bus.

Solution is explanation in proxy.h. Diff for snapper.cc shows how the proxy classes replace the no_dbus-if.

The code is not finished, it is just a prototype. Not all snapper commands are based on the new classes.

Two approvals required.

### DIFF
--- a/client/proxy-dbus.cc
+++ b/client/proxy-dbus.cc
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2016 SUSE LLC
+ *
+ * All Rights Reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of version 2 of the GNU General Public License as published
+ * by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, contact Novell, Inc.
+ *
+ * To contact Novell about this file by physical or electronic mail, you may
+ * find current contact information at www.novell.com.
+ */
+
+
+#include <sstream>
+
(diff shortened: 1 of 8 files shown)


ancorgs:
From the OOP point of view (taking into account my limited C++ knowledge) this looks ok. Enough to consider the PBI as done from my POV. Now we need a couple of C++-capable eyes to confirm I not overlooking anything language-specific.

With no intention to start a flame, I cannot resist to say that this will be way easier to implement in Ruby thanks to duck typing and using modules like `Forwardable`, `Delegator` and `SimpleDelegator`. :wink:

client/proxy-dbus.cc
```diff
@@ -0,0 +1,150 @@
...
+
+ProxySnapshots::const_iterator
+ProxySnapperDbus::createSingleSnapshot(const SCD& scd)
+{
+    unsigned int num = command_create_single_xsnapshot(*backref->conn, config_name, scd.description,
+						       scd.cleanup, scd.userdata);
+
+    proxy_snapshots.emplace_back(new ProxySnapshotDbus(&proxy_snapshots, num));
+
+    return --proxy_snapshots.end();
```
shundhammer: When I first saw this, it looked somehow wrong to me: Take the iterator that points one element beyond the last element of the container and go one element back.

But it looks like STL containers really can't do that any better: vector::back() returns a reference to the last element, not an iterator; vector::rbegin() returns a reverse iterator, not a forward iterator as everybody using such a function would expect. So this is indeed the best solution STL has to offer.

shundhammer:
In some C++ based frameworks like for example Qt, you can also use delegate classes in certain cases. C++ also offer templates, but they tend to produce nightmarish code.

Somehow I have the feeling it should be easier and less repetitive to do an abstraction layer like those proxy classes, but I can't quite put my finger on it, and I don't have a good idea how to do that from the top of my head.

But since at least the amount of functions that are affected is not that bad, I tend to say let's go for this approach. Maybe we can come up with a better solution some day.

Until that day, LGTM.

PR #299, opened 2016-12-13
support option --no-dbus for all snapper commands

aschnell:
For https://trello.com/c/w79xWag5/461-5-fate-321049-make-snapper-create-work-without-dbus.


client/proxy-dbus.cc
```diff
@@ -0,0 +1,360 @@
...
+    cleanup = x.cleanup;
+    userdata = x.userdata;
+}
+
+
+ProxySnapshotDbus::ProxySnapshotDbus(ProxySnapshotsDbus* backref, SnapshotType type, unsigned int num,
+				     time_t date, uid_t uid, unsigned int pre_num,
+				     const string& description, const string& cleanup,
+				     const map<string, string>& userdata)
+    : backref(backref), type(type), num(num), date(date), uid(uid), pre_num(pre_num),
```
shundhammer: If you write that one element per line, it's very easy to compare it against the data members in the header file. Like this, that's very tedious.

client/proxy.cc
```diff
@@ -0,0 +1,188 @@
...
+ProxySnapshots::findNum(const string& str)
+{
+    istringstream s(str);
+    unsigned int num = 0;
+    s >> num;
+
+    if (s.fail() || !s.eof())
+    {
+	cerr << sformat(_("Invalid snapshot '%s'."), str.c_str()) << endl;
+	exit(EXIT_FAILURE);
```
shundhammer: exit() ? is't that a bit drastic? Why not throw an exception and let the caller decide how to handle that?

client/proxy.cc
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    iterator ret = find(num);
+    if (ret == proxy_snapshots.end())
+    {
+	cerr << sformat(_("Snapshot '%u' not found."), num) << endl;
+	exit(EXIT_FAILURE);
```
shundhammer: How about returning an end() iterator here?

While the above exit() can be explained with an I/O error that is hard to recover from, this might be a simple user error (typo, for example). If one day there is something like a snapper shell (you'll never know), this error could be presented to the user so he can fix it and try with another number. Calling exit() immeditaly here would make this impossible.

client/proxy.cc
```diff
@@ -0,0 +1,188 @@
...
+    {
+	cerr << sformat(_("Invalid snapshot '%s'."), str.c_str()) << endl;
+	exit(EXIT_FAILURE);
+    }
+
+    const_iterator ret = find(num);
+    if (ret == proxy_snapshots.end())
+    {
+	cerr << sformat(_("Snapshot '%u' not found."), num) << endl;
+	exit(EXIT_FAILURE);
```
shundhammer: Again: end() iterator might be preferable.

shundhammer:
Besides the small things mentioned, LGTM.

client/proxy.cc
aschnell: In general exit() is drastic. But these functions here are only used to parse the command line and any error must be shows to the user and abort the program anyway.

mvidner:
The last commits (PR #307) LGTM.

Just a closing paren missing in the changelog.

PR #309, opened 2016-12-21
- fixed --root option (bsc#1016633)

aschnell:


client/proxy-lib.h
```diff
@@ -86,8 +86,8 @@ class ProxySnapperLib : public ProxySnapper
 
 public:
 
-    ProxySnapperLib(const string& config_name)
-	: snapper(new Snapper(config_name, "/")), proxy_snapshots(this)
+    ProxySnapperLib(const string& config_name, const string& target_root)
```
jreidinger: this is just some internal class? I miss any documentation here. At least link to what it proxy.

jreidinger:
in general I miss testcase as it is very often used option in yast upgrade, I will welcome to have at least small trivial test.

aschnell:
Yes, integration tests are missing as I mentioned in the call. Many are needed: 19 commands, --no-dbus, --root -> that makes already 57 tests. Additional many commands have several major modes, e.g. two for rollback, 2 for cleanup, 4 for create, ... And I'm not even talking about tests for wrong usage.

But so far always manual testing was considered enough by the PO.

client/proxy-lib.h
aschnell: Yes, it is a internal class. The concept is explained in proxy.h.

jreidinger:
ok, so LGTM

Issue #268, opened 2016-09-08
Maximum description length for snapper


NicoHood:
If you call `snapper list` and have snapshots with a very long description it will produce an unreadable table. This bug was noticed with the pacman hooks of snap-pac: https://github.com/wesbarnett/snap-pac/issues/4

It would be nice to cut off the table at a maximum number/terminal width and use `--full` for the full output if desired.

kmotavalli:
https://github.com/kmotavalli/snapper-1/pull/3/commits

something like this?

aschnell:
With https://github.com/openSUSE/snapper/pull/612 there is a new option ```--abbreviate```. It is not the default yet.

PR #316, opened 2017-01-11
experimental: add list --full option, truncate description by default

kmotavalli:
trying to implement the option required by @NicoHood in #268.

I have not tested nor tried to compile it. I'm not even sure if the logic is right.


kmotavalli:
wops, i'll fix the commits.

Issue #310, opened 2016-12-22
Please do not mess debian/ directory


henrich:
From Debian package maintainer's view, I don't want upstream to put files into debian/ directory.
It very confuse me. Could you consider to rename and handle it as other directory name?

lslezak:
This has been fixed in #319 - the directory has been moved to `dists/debian` to not conflict with the distribution packaging.

henrich:
thanks!😃

PR #319, opened 2017-01-13
Improve Debian Packaging

lslezak:
- Build a single tarball with all Debian packaging (workaround for an OBS bug   https://github.com/openSUSE/obs-service-source_validator/pull/41 which caused automatic SR rejection, see [this SR](https://build.opensuse.org/request/show/447293)).
- This also makes the packaging less messy in OBS (less files).
- Rename `debian/` to `package_deb/` to not conflict with upstream packaging (#310).
- Testing builds verified at [home:lslezak:branches:filesystems:snapper](https://build.opensuse.org/project/show/home:lslezak:branches:filesystems:snapper).


aschnell:
Is "package_dep" commonly used or why do you use an abbreviation?

PR #323, opened 2017-02-03
- deal with CaaSP btrfs setup where certain subvolumes are read-only

aschnell:
For https://trello.com/c/3j0lyxqY/497-2-casp-1018095-snapper-cleanup-failure and https://trello.com/c/nvR379s2/512-2-bsc-1018302-rollback-does-not-work-because-of-ro-filesystem.


snapper/Btrfs.cc
```diff
@@ -1446,19 +1453,19 @@ namespace snapper
...
 	try
 	{
+	    SDir general_dir = openGeneralDir();
+
 	    if (num == 0)
 	    {
 		SDir subvolume_dir = openSubvolumeDir();
 		subvolid_t id = get_id(subvolume_dir.fd());
-		set_default_id(subvolume_dir.fd(), id);
+		set_default_id(general_dir.fd(), id);
```
joseivanlopez: Just out of curiosity, why some methods like `set_default_id` are with underscores and other ones with camel case?

aschnell: In C++ there is no single coding style so sometimes they are unfortunately mixed.

PR #324, opened 2017-02-06
Run the osc source validator

lslezak:
Run the osc source validator in the openSUSE based Travis builds.

Similar to the https://github.com/yast/docker-yast-cpp/pull/1 and https://github.com/yast/docker-yast-ruby/pull/2 pull requests, run the OBS source validator also here.

I found out that in [this YaST PR](https://github.com/yast/yast-packager/pull/225) build succeeded at Travis but [failed in Jenkins](https://ci.opensuse.org/job/yast-packager-github-push/231//console). It turned out that the Jenkins failure was caused by a wrong changelog sequence (dates not in ascending order) which was not tested at Travis.

This is the same enhancement also for snapper.

### DIFF
--- a/.travis.tumbleweed.sh
+++ b/.travis.tumbleweed.sh
@@ -5,6 +5,9 @@ set -e -x
 make -f Makefile.repo
 make package
 
+# run the osc source validator to check the .spec and .changes locally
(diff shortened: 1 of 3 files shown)


shundhammer:
LGTM

Issue #330, opened 2017-03-06
Debian: Packaging issue causing timeline snapshots to not run.


ttelford:
I noticed my system wasn't making `timeline` snapshots for the past two months.

I tried [reinstalling the package](https://build.opensuse.org/package/show/filesystems:snapper/snapper), and saw the following error:

    Unpacking snapper (0.4.3) over (0.4.3) ...
    dpkg: warning: snapper: conffile '/etc/cron.daily/snapper' is not a plain file or symlink (= '/etc/cron.daily/snapper')
    dpkg: warning: snapper: conffile '/etc/cron.hourly/snapper' is not a plain file or symlink (= '/etc/cron.hourly/snapper')

It appears that the package tries to create four files:

    /etc/cron.daily/snapper
    /etc/cron.daily/snapper/suse.de-snapper
    /etc/cron.hourly/snapper
    /etc/cron.hourly/snapper/suse.de-snapper

Note these are *files* - `/etc/cron.daily/snapper` can't be both a file and a directory; and the same is true of `/etc/cron.hourly/snapper`.

So when unpacked, directories are created (`/etc/cron.hourly/snapper` and `/etc/cron.daily/snapper`), and since it can't overwrite the directory with a file, `dpkg` appears to fall back to rename the `/etc/cron.(hourly|daily)/snapper` script to `/etc/cron.{hourly|daily)/snapper.dpkg-new`.

The end result is that there are actually _two_ copies of the same cron job in `/etc/cron.daily` and `/etc/cron.hourly`, and because of their naming/placement, neither is executed.

I resolved it by deleting the `/etc/cron.daily/snapper` directory and renaming `/etc/cron.daily/snaper.dpkg-new` to `/etc/cron.daily/snapper` (and the same for `/etc/cron.hourly`)

lslezak:
Oh, by mistake the cron files were installed in a wrong location (into a subdirectory), this has been fixed in #329.

However, at the package upgrade there is a conflict between directory (old broken version) and a file (new fixed version). We need a preinst script to handle to specific upgrade....

lslezak:
Should be fixed in the latest version by #332, just wait a bit for the OBS to build and publish the fixed packages and then try upgrading again.

PR #332, opened 2017-03-09
Debian: Added snapper.preinst

lslezak:
- This script correctly handles upgrade from the previously broken version.
- Fixes issue #330 
- Tested manually in a Docker instance


aschnell:
LGTM

PR #339, opened 2017-03-23
Fix broken video link on snapper.io homepage

sielicki:
The link was broken. This broken link was fixed in videos.html, but never made it into index.html


aschnell:
Thanks, LGTM.

PR #346, opened 2017-05-09
- set cleanup algorithm for rollback snapshots (fate#321773)

aschnell:
For https://trello.com/c/ExzdxWSN/587-3-fate-321773-cleanup-of-old-rootfs-after-rollback.


client/snapper.cc
```diff
@@ -1203,11 +1212,16 @@ command_rollback(ProxySnappers* snappers, ProxySnapper* snapper)
...
+	cerr << sformat(_("Command 'rollback' cannot be used on a non-root subvolume %s."),
 			subvolume.c_str()) << endl;
-        exit(EXIT_FAILURE);
+	exit(EXIT_FAILURE);
     }
 
+    pair<bool, unsigned int> previous_default = filesystem->getDefault();
+
+    if (previous_default.first && scd1.description == "rollback backup")
+	scd1.description += sformat(" of #%d", previous_default.second);
```
shundhammer: No  _("...") translation marker here?

Also, adding strings is something that gets broken easily with translations.

aschnell: No, other descriptions also seem to be not translated, e.g. https://github.com/yast/yast-storage/blob/master/src/modules/StorageSnapper.rb#L82.

shundhammer: OK then.

shundhammer:
LGTM

PR #351, opened 2017-08-03
Fixed table formatting for non-ACSII characters (bsc#1051901)

lslezak:
## The Issue

- This patch fixes bug [bsc#1051901](https://bugzilla.suse.com/show_bug.cgi?id=1051901) - printing  a broken table 
- A similar problem was reported in zypper, see [bsc#950869](https://bugzilla.suse.com/show_bug.cgi?id=950869).
- The problem is that the `::mbrtowc()` behavior has been changed in glibc-2.22, see [the patch](
http://sourceware.org/git/?p=glibc.git;a=commit;h=9781a37)
- Now the function returns `-2` error when `n = 0`, the column width computation always failed and the code used the raw byte size as the fallback

## The Fix

- The fix is backward compatible and works with the old glibc as well, the only difference is that the loop is not finished by the `::mbrtowc()` call processing the terminating `NULL`, but sooner when the end of the string is detected.
- Version 0.5.1

### The Original Output

```
Typ    | #  | Před # | Datum                                 | Uživatel | Úklid | Popis                 | Uživatelská data
-------+----+---------+---------------------------------------+-----------+--------+-----------------------+-------------------
single | 0  |         |                                       | root      |        | current               |                   
single | 1  |         | Čt 3. srpen 2017, 16:02:46 CEST | root      |        | first root filesystem |                   

```

### With the Fix

```
Typ    | #  | Před # | Datum                           | Uživatel | Úklid  | Popis                 | Uživatelská data
-------+----+--------+---------------------------------+----------+--------+-----------------------+-----------------
single | 0  |        |                                 | root     |        | current               |                 
single | 1  |        | Čt 3. srpen 2017, 16:02:46 CEST | root     |        | first root filesystem |                 
```

```

```


aschnell:
Thanks. LGTM

lslezak:
Travis fails for Tumbleweed because of `cannot find -ldbus-1` error. I have the same problem on my machine running TW so it seems TW is somehow broken. The other distributions build correctly so lets ignore that.

PR #352, opened 2017-08-12
spec: trim parts unwanted by openSUSE

jengelh:
Remove author list, remove redundant %clean section and other ancient
rpm lines, remove unneeded .la files, and remove the Prefix: tag
which is practically not supported by any distro. Replace old $RPM_*
variables by macros.


aschnell:
LGTM

PR #374, opened 2017-12-05
Replace references to /var/adm/fillup-templates with new %_fillupdir …

sysrich:
…macro (boo#1069468)


aschnell:
Should be fine.

PR #380, opened 2018-01-10
Update the distro versions used at Travis

lslezak:
## Summary 
- Use newer distros at Travis
- I do not think we should still build for the older ones, we should focus on the newer versions. But if needed the older versions can be added and built in parallel...

### Version Updates
- Debian 8 -> 9
- Fedora 25 -> 27
- Ubuntu 16.10 -> 17.10


shundhammer:
LGTM

aschnell:
Also LGTM

PR #381, opened 2018-01-15
Build reproducible tarballs by "make package"

lslezak:
- `make package` now builds reproducible tarballs (1:1 bit copy with the same checksum)
- Makes the Open Build Service submissions easier (the internal and external OBS will contain the same tarballs regardless where/when they are build)
- Similar solution as used in https://github.com/openSUSE/packaging_rake_tasks/pull/43 and https://github.com/openSUSE/packaging_rake_tasks/pull/44

Runnig `make package` now always produces the same result:

```console
> make package > /dev/null
> sha1sum package/snapper-0.5.3.tar.bz2 package/debian.tar.gz 
02a9eeb9447eb0ea8de56ff549a48d165d71012f  package/snapper-0.5.3.tar.bz2
ddc6b7e289a95fa7ae9cc491d403470e423b21ea  package/debian.tar.gz
> rm package/snapper-0.5.3.tar.bz2 package/debian.tar.gz 
> make package > /dev/null
> sha1sum package/snapper-0.5.3.tar.bz2 package/debian.tar.gz 
02a9eeb9447eb0ea8de56ff549a48d165d71012f  package/snapper-0.5.3.tar.bz2
ddc6b7e289a95fa7ae9cc491d403470e423b21ea  package/debian.tar.gz

```

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -60,9 +60,32 @@ package-clean:
 # VERSION file.
 #
 # $< is the first depencency of the rule, i.e. snapper-$(VERSION).tar.gz in this case.
-#
+
+# Build a reproducible tarball:
+# - set the file time stamps according to the latest commit
+# - sort the files (in a locale independent way, use the NULL separator to
+#   correctly process also the file names containing a new line)
+#   Note: tar >= 1.28 supports "--sort=name" option, unfortunately
+#   Leap 42.3 and SLES12-SP3 contain version 1.27.1
+# - use the GNU format (the default POSIX format contains some time stamps)
+# - set the owner and group to "root"
+# - set the fixed modification time
+
+# shared tar options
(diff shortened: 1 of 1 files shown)


aschnell:
LGTM

But in general I wonder whether the approach of the CI is suboptimal if the standard automake way is not working for the CI. Also generating the tar-ball from the extracted tar-ball (so without git) is not supported anymore AFAIS.

lslezak:
I have tested generating tarball from the tarball and it works (no crash). The only drawback that the file time stamps are not set correctly as the git commit date is not available. It's set to today midnight (00:00) so the build is reproducible during the same day. That's acceptable IMHO.

lslezak:
BTW the plan is to improve the OBS integration later...

PR #387, opened 2018-01-28
Command was wrongly translated

IsaakGroepT:
Someone translated the command "snapper create" wrongly to Dutch. I undid that


aschnell:
Please use Weblate as explained in https://github.com/openSUSE/snapper/blob/master/po/README.

IsaakGroepT:
I'd rather not go to the unnecessary process of creating an Opensuse account and trying to decipher the use of Weblate. In any case, I've done my part.

PR #388, opened 2018-01-29
systemd: Make timeline.timer persistent

mrkline:
Without this, it, timeline snapshots might not be taken for up to 59
minutes after boot. This is troublesome for laptops and other use cases
where the user might not have their machine on for that long.

On a less pressing, but related note,  why doesn't `cleanup.timer` run for ten minutes after system startup? (If the main concern is to make sure `boot.timer` has been run before, wouldn't a minute or two suffice?)


aschnell:
I cannot remember why the cleanup.timer is setup that way. Maybe to avoid extra IO during boot.
But boot.timer was added years later so that cannot be the reason.

mrkline:
Checking in a year later - @aschnell, are there objections to this change? Is there somewhere else I should try to get it upstream?

PR #391, opened 2018-02-05
Use 'idle' IO and cpu scheduling class for cleanup service to prevent slowdown of systems

okurz:
Related to https://bugzilla.suse.com/show_bug.cgi?id=1063638 . Not only btrfs
maintenance tasks can bring systems to a halt or at least slow down and they
already set idle a class.

### DIFF
--- a/data/cleanup.service
+++ b/data/cleanup.service
@@ -6,4 +6,5 @@ Documentation=man:snapper(8) man:snapper-configs(5)
 [Service]
 Type=simple
 ExecStart=/usr/lib/snapper/systemd-helper --cleanup
-
+IOSchedulingClass=idle
+CPUSchedulingPolicy=idle


aschnell:
LGTM

But from my experience I doubt it will change much.

okurz:
Wow, that was fast :) yes, I also think it won't solve a lot

fcrozat:
I'm commenting on this old PR because it doesn't seem to do what we expect it to do:
systemd-helper is properly run with the idle IO class and idle niceness but the btrfs cleanup kernel thread is still running with best effort. Does the systemd-helper initiate the delete snapshot call itself to the kernel or does it send a dbus call to snapperd, which does it ? If this is the second option, it would explain why snapshot deletion can still slow down a system a lot (I'm seeing it on 15SP2).

aschnell:
It is the second option. Also see the discussion at https://github.com/openSUSE/snapper/pull/456.

PR #403, opened 2018-04-20
- soften error handling if setting locale fails due to broken system setup (bsc#1085832)

aschnell:
For https://trello.com/c/RYVaFyPX/351-2-opensuse-p2-1085832-snapper-does-not-work-without-locale-installed.

### DIFF
--- a/client/snapper.cc
+++ b/client/snapper.cc
@@ -1522,7 +1522,6 @@ main(int argc, char** argv)
     catch (const runtime_error& e)
     {
 	cerr << "Failed to set locale. Fix your system." << endl;
-	exit(EXIT_FAILURE);
     }
 
     setLogDo(&log_do);
--- a/client/snapper.cc
+++ b/client/snapper.cc
@@ -1522,12 +1522,12 @@ main(int argc, char** argv)
     }
     catch (const runtime_error& e)
     {
-	cerr << "Failed to set locale. Fix your system." << endl;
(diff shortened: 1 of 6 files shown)


client/snapper.cc
```diff
@@ -1522,7 +1523,11 @@ main(int argc, char** argv)
     catch (const runtime_error& e)
     {
 	cerr << "Failed to set locale. Fix your system." << endl;
-	exit(EXIT_FAILURE);
+    }
+
+    if (strcmp(nl_langinfo(CODESET), "UTF-8") != 0)
+    {
+	cerr << "Running in non UTF-8 locale. Setup is unsupported." << endl;
```
lslezak: Q: The STDERR messages are not translated in snapper? E.g. the `--help` output few lines above is translated, but that goes to STDOUT...

lslezak: Actually this STDERR https://github.com/openSUSE/snapper/pull/403/files#diff-b5c274f368d2fab68ee1bc4f9206df4eR1490 is translated, so I guess these two should be as well...

aschnell: AFAIS the first will never be translated since setting the locale failed. Anyway, marked to be translated.

jsrain:
Not sure why the .pot-file is now in git (as it can be generated), but LGTM (assuming pending checks succeed)

aschnell:
Weblate needs the pot file. But still Weblate cannot merge the po files. Sad but true.

PR #415, opened 2018-07-23
- fixed logging during shutdown

aschnell:
For https://trello.com/c/USeFyjd0/100-sle15-p2-1079385-snapperd-dumped-core-with and others.


okurz:
+1 but the git commit message subject lines with leading '-' look unusual

server/MetaSnapper.cc
```diff
@@ -198,6 +200,14 @@ MetaSnappers::init()
 }
 
 
+void
+MetaSnappers::unload()
+{
+    for (iterator it = entries.begin(); it != entries.end(); ++it)
```
lslezak: NP: I'd prefer the new syntax without iterators and increment:
```c++
for(auto& entry : entries)
    entry.unload();
```

snapper/Logger.cc
```diff
@@ -37,10 +38,25 @@ namespace snapper
...
-    // Use a pointer to avoid a global destructor. Otherwise the Snapper
-    // destructor in Factory.cc can be called after when logging does not work
-    // anymore. TODO: nicer solution.
-    string* filename = new string("/var/log/snapper.log");
+    namespace
+    {
+
+	struct LoggerData
+	{
+	    LoggerData() : filename("/var/log/snapper.log"), mutex() {}
```
lslezak: NP: This string is hardcoded in several places, it would be nice to have a constant for it to have a single place to change. BTW the log file is not configurable?

snapper/Logger.cc
```diff
@@ -69,11 +85,9 @@ namespace snapper
 	string prefix = sformat("%s %s libsnapper(%d) %s(%s):%d", datetime(time(0), false, true).c_str(),
 				ln[level], getpid(), file, func, line);
 
-	static boost::mutex mutex;
-
-	boost::lock_guard<boost::mutex> lock(mutex);
+	boost::lock_guard<boost::mutex> lock(logger_data->mutex);
 
-	FILE* f = fopen(filename->c_str(), "ae");
+	FILE* f = fopen(logger_data->filename.c_str(), "ae");
```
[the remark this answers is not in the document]
aschnell: There is ofsteam. But "e" flag (O_CLOEXEC) cannot be provided there AFAIR.

server/MetaSnapper.cc
aschnell: Snapper still uses the old syntax since it was created mainly before C++11.

snapper/Logger.cc
aschnell: Added a define.

No, the log filename is not configurable. You would also need to modify logrotate in that case so it is not easy.

snapper/Logger.cc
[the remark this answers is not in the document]
lslezak: OK, it seems C++ does not support `O_CLOEXEC`... :worried:

server/MetaSnapper.cc
lslezak: OK, but could we use the new syntax while we are touching the code anyway? In YaST we also still have a lot of YCP-like Ruby code but we are gradually rewriting it to pure Ruby when touching it. Would this approach work also for snapper?

aschnell: OK, I have changed it to the new C++11 syntax. But given the current activity of the snapper project there will be a mixture of syntaxes for many years.

PR #427, opened 2018-09-14
- avoid setenv after fork (bsc#1107587)

aschnell:


snapper/SystemCmd.cc
```diff
@@ -231,8 +235,7 @@ SystemCmd::doExecute( const string& Cmd )
 		    y2err("close child failed errno:" << errno << " (" << stringerror(errno) << ")");
 		    }
 		closeOpenFds();
-		Ret_i = execl( Shell_Ci.c_str(), Shell_Ci.c_str(), "-c",
-			       Cmd.c_str(), NULL );
+		Ret_i = execle(Shell_Ci.c_str(), Shell_Ci.c_str(), "-c", Cmd.c_str(), nullptr, &env[0]);
```
lslezak: I'm not a C++ expert, is it OK to use the address of the first item in a C++ vector as a C array?

aschnell: Yes. see e.g. https://en.cppreference.com/w/cpp/container/vector:

> The elements are stored contiguously, which means that elements can be accessed not only through iterators, but also using offsets to regular pointers to elements. [...]

PR #433, opened 2018-10-05
- right align used space in table

aschnell:
For https://trello.com/c/Mm12Zg7N/296-fate-323843-show-unique-space-consumed-by-snapshot-in-snapper-list.


client/utils/Table.cc
```diff
@@ -111,15 +105,25 @@ void TableRow::dumpTo (ostream &stream, const Table & parent) const
     }
     else
     {
-      stream << s;
-      stream.width (parent._max_width[c] - ssize);
+      if (parent._header.align(c) == TableAlign::LEFT)
+	stream << s << setw(parent._max_width[c] - ssize) << "";
+      else
+	stream << setw(parent._max_width[c] - ssize) << "" << s;
```
jreidinger: I am not sure if it is bug in github, but indentation looks strange here.

aschnell: Somehow github does not handle the spaces and tabs correct.

mvidner: FYI, GitHub assumes the Tab character has stops each 4 columns while the text terminals assume 8 columns. So it's best not to mix tabs and spaces in files.

PR #435, opened 2018-10-11
- mark active and default btrfs subvolume in list output (fate#326479)

aschnell:
For https://trello.com/c/NVsBNefc/293-fate-326479-snapper-list-mark-current-and-active-snapshot-and-prevent-deleting-them.


shundhammer:
LGTM

PR #437, opened 2018-10-12
Ensure also "timeline" runs with 'idle' IO and CPU scheduling {boo#1099494)

okurz:
With https://github.com/openSUSE/snapper/pull/391 we already asked the
background "cleanup" service to take idle IO and CPU priority so for
sake of consistency and potentially fix bad responsiveness of systems
also the timeline service should be triggered the same.

This change is motivated by an observation I made on my own system: The
timeline service was triggered at around the same time a personal sync
and backup script was triggered. Then the system turned unresponsive for
some minutes. I guess asking the timeline service to act with idle
priority might help a bit in similar situations.

### DIFF
--- a/data/timeline.service
+++ b/data/timeline.service
@@ -6,4 +6,6 @@ Documentation=man:snapper(8) man:snapper-configs(5)
 [Service]
 Type=simple
 ExecStart=/usr/lib/snapper/systemd-helper --timeline
+IOSchedulingClass=idle
+CPUSchedulingPolicy=idle
 


aschnell:
LGTM

But I doubt this will have any effect. For once creating a single snapshot (that is what the timeline service does) should not cause much IO. On the other hand the systemd-helper just tells snapperd to do some actions and snapperd is unaffected by the change.

okurz:
hm, maybe. So can we ensure that snapperd itself runs with less aggressive IO class?

aschnell:
I would not call the default "best-effort" IO class aggressive - "realtime" would be.

Anyway, setting the IO class for all of snapperd does not look so good to me so it would have to be per task. That looks like a huge effort.

okurz:
Discussed with @aschnell in person. I agree that setting the scheduling class in the service definition file probably has no effect at all as the important process is snapperd which is triggered over dbus so the scheduling prios would not be taken over. I will experiment with calling snapperd directly with something like `nice -n 19 ionice -c 3 snapperd…" but I agree that this is most likely not be what we want in all cases for all tasks.

PR #447, opened 2018-10-29
- extended space aware cleanup

aschnell:
For https://trello.com/c/iePJYmUA/294-fate-325774-enhanced-snapper-space-aware-clean-up-algorithm-to-ensure-minimal-free-space-is-available.


client/cleanup.cc
```diff
@@ -223,41 +236,80 @@ Cleaner::is_quota_aware() const
...
+    return parameters.space_limit < 1.0;
+}
+
+
+bool
+Cleaner::is_quota_satisfied() const
+{
+    QuotaData quota_data = snapper->queryQuotaData();
+
+    double fraction = (double)(quota_data.used) / (double)(quota_data.size);
```
shundhammer: Check for `quota_data.size != 0` to prevent SIGDIV?

aschnell: Yes, in theory possible. BTW fraction would likely be ```inf``` in that case. I will add a check.

PR #456, opened 2018-11-27
[WIP] Ensure snapperd runs with 'idle' IO and CPU scheduling (boo#1099494)

okurz:
With https://github.com/openSUSE/snapper/pull/391 we already asked the
background "cleanup" service to take idle IO and CPU priority so for
sake of consistency and potentially fix bad responsiveness of systems
the snapperd service which is triggered over dbus should be asked to run
in a 'nice' way.

This change is motivated by an observation I made on my own system: The
timeline service was triggered at around the same time a personal sync
and backup script was triggered. Then the system turned unresponsive for
some minutes. I guess asking the snapperd service to act with idle
priority might help a bit in similar situations.

Changing the the timeline or cleanup service does not help as they would
only trigger snapperd and not produce significant IO themselves and the
nice levels are not communicated to snapperd otherwise.

I have been running this change for some days now and did not observe
any problems. I could ensure that snapperd would still start up and
trigger the necessary actions and I have not observed any negative
impact so it might fix the original issue but this is not ensured.

### DIFF
--- a/data/cleanup.service
+++ b/data/cleanup.service
@@ -6,5 +6,3 @@ Documentation=man:snapper(8) man:snapper-configs(5)
 [Service]
 Type=simple
 ExecStart=/usr/lib/snapper/systemd-helper --cleanup
-IOSchedulingClass=idle
-CPUSchedulingPolicy=idle
--- a/data/org.opensuse.Snapper.service
+++ b/data/org.opensuse.Snapper.service
@@ -1,5 +1,5 @@
 # DBus service activation config
 [D-BUS Service]
 Name=org.opensuse.Snapper
-Exec=/usr/sbin/snapperd
+Exec=/usr/bin/nice /usr/bin/ionice -c 3 /usr/sbin/snapperd
 User=root


aschnell:
With this change even snapper operations requested by the user (e.g. interactive on the command line) run with low priority and can liekly take way longer than so far. This might been seen as a regression. It might also trigger more problem like https://bugzilla.suse.com/show_bug.cgi?id=1099494.

okurz:
ah, right. When creating the PR I completely forgot one think I wanted to ask feedback for: How to make that configurable. Any ideas?

aschnell:
That is likely difficult: The client would need an option for the "priority" and pass that to the server. The server would have to handle it (which is not so easy since, e.g. the server has a queue of comparisons to make which would also need to be sorted by priority and getting a high priority request would have to interrupt a low priority request. Setting all those nice levels in a multi-threaded program might also be difficult).

okurz:
ok, so how about we aim for a compromise and schedule with a still "nice" I/O level, but not idle? The only other alternative I see is wait for the Linux kernel to change behaviour and handle any background tasks to be of always low/no impact to any foreground task but I see that unlikely to happen

aschnell:
Also only using nice I/O level can slow down interactive snapper commands by background tasks.

okurz:
I see, so interactive snapper commands trigger snapperd over dbus but wait synchronously for snapperd to finish tasks, right?

aschnell:
Yes, snapper waits for the result from snapperd.

okurz:
ok, how about the following:

* Not change the existing dbus service definition but add another explicit one with the nice levels
* Add an option to snapper, e.g. "--nice" to trigger either the old service or the "nice" one
* Call snapper with the option "--nice" for only the background services, e.g. cleanup, but not the interactive calls

aschnell:
That could lead to problems when snapperd gets started twice (which could happen then).

As I already wrote the best approach should be that snapperd is informed about the priority
and handles it.

fcrozat:
@aschnell should I open a feature request on SUSE side, so you can have some dedicated time to work on it ?

aschnell:
Yes, a feature request should be opened. I expect it to include the problem to solve and some evidence that setting priorities does indeed solve the problem.

PR #458, opened 2018-11-28
Add option to disable qgroup rescan when listing snapshots

gergo-b:
Currently listing snapshots of a config with sizes involves complete qgroup rescan of the underlying btrfs file system. This means that listing multiple configs from the same file system, the same file system is rescanned several times.

I've added an option to the `list` command to skip this explicit rescanning. So the listing is faster, but might show outdated information.

As a next step I thought that instead of requesting the rescans individually for each config, it would be better to request rescan for a list of configs and the processing side should ensure that each involved btrfs file system is rescanned exactly once. Any ideas/comments?

### DIFF
--- a/client/snapper.cc
+++ b/client/snapper.cc
@@ -452,6 +452,11 @@ list_from_one_config(ProxySnapper* snapper, ListMode list_mode, bool show_used_s
     {
 	default_snapshot = snapshots.getDefault();
 	active_snapshot = snapshots.getActive();
+
+	if (show_used_space)
+	{
+	    show_used_space = snapper->isQuotaEnabled();
+	}
     }
     catch (const DBus::ErrorException& e)
     {
--- a/client/snapper.cc
+++ b/client/snapper.cc
@@ -356,6 +356,7 @@ help_list()
 	 << _("    Options for 'list' command:") << '\n'
 	 << _("\t--type, -t <type>\t\tType of snapshots to list.") << '\n'
 	 << _("\t--disable-used-space\t\tDisable showing used space.") << '\n'
+	 << _("\t--no-used-space-update\t\tDo not update used space information.") << '\n'
 	 << _("\t--all-configs, -a\t\tList snapshots from all accessible configs.") << '\n'
 	 << endl;
 }
@@ -364,7 +365,7 @@ enum ListMode { LM_ALL, LM_SINGLE, LM_PRE_POST };
 
 
 void
-list_from_one_config(ProxySnapper* snapper, ListMode list_mode, bool show_used_space);
+list_from_one_config(ProxySnapper* snapper, ListMode list_mode, bool show_used_space, bool update_used_space);
 
 
 void
@@ -373,6 +374,7 @@ command_list(ProxySnappers* snappers, ProxySnapper*)
     const struct option options[] = {
 	{ "type",		required_argument,	0,	't' },
 	{ "disable-used-space", no_argument,            0,      0 },
+	{ "no-used-space-update",no_argument,		0,	0 },
 	{ "all-configs",	no_argument,		0,	'a' },
 	{ 0, 0, 0, 0 }
     };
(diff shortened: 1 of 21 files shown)


client/snapper.cc
```diff
@@ -356,6 +356,7 @@ help_list()
 	 << _("    Options for 'list' command:") << '\n'
 	 << _("\t--type, -t <type>\t\tType of snapshots to list.") << '\n'
 	 << _("\t--disable-used-space\t\tDisable showing used space.") << '\n'
+	 << _("\t--no-qgroup-rescan\t\tDo not rescan groups.") << '\n'
```
aschnell: Please use a generic name, e.g. no-used-space-rescan. snapper is not only about btrfs and other storage systems/file systems might also need that option.

client/snapper.cc
```diff
@@ -468,7 +476,7 @@ list_from_one_config(ProxySnapper* snapper, ListMode list_mode, bool show_used_s
     if (list_mode != LM_ALL && list_mode != LM_SINGLE)
 	show_used_space = false;
 
-    if (show_used_space)
+    if (show_used_space && rescan_qgroups)
```
aschnell: AFAIS this breaks for btrfs if quota is not enabled or LVM since the try/catch is also used to disable show_used_space if it is not supported. So using the new option will cause an error in these cases.

client/snapper.cc
gergo-b: I've renamed to option to be generic and consistent with the disable option and documented it in the referenced file.

client/snapper.cc
gergo-b: Yes, indeed it broke in those cases. Instead of abusing some other method to force disable the used space display, I've extended the Snapper interface to explicitly check for quota support.

snapper/Snapper.cc
```diff
@@ -655,6 +655,16 @@ namespace snapper
 	    SN_THROW(AclException());
     }
 
+    bool
+    Snapper::isQuotaSupported() const
+    {
+#ifdef ENABLE_BTRFS_QUOTA
+	const Btrfs* btrfs = dynamic_cast<const Btrfs*>(getFilesystem());
+	return btrfs != nullptr;
```
aschnell: This does not look like good C++. A virtual function in Filesystem overridden in Btrfs would be much cleaner.

server/Client.cc
```diff
@@ -320,6 +320,11 @@ Client::introspect(DBus::Connection& conn, DBus::Message& msg)
 	"      <arg name='number' type='u' direction='out'/>\n"
 	"    </method>\n"
 
+	"    <method name='IsQuotaSupported'>\n"
+	"      <arg name='config-name' type='s' direction='in'/>\n"
+	"      <arg name='is_quota_supported' type='b' direction='out'/>\n"
```
aschnell: All other method names use '-' instead of '_'.

client/snapper.cc
```diff
@@ -452,6 +460,11 @@ list_from_one_config(ProxySnapper* snapper, ListMode list_mode, bool show_used_s
     {
 	default_snapshot = snapshots.getDefault();
 	active_snapshot = snapshots.getActive();
+
+	if (show_used_space)
+	{
+	    show_used_space = snapper->isQuotaSupported();
```
aschnell: AFAIS this will still not work if quota is disabled on btrfs. show_used_space will still be true and the getUsedSpace() calls will fail.

snapper/Snapper.cc
gergo-b: I've extended the class Filesytem to check if quota is enabled.

client/snapper.cc
gergo-b: Now it checks whether quota is enabled or not. So in case of Btrfs with disabled quota the `getUsedSpace()` won't be called. Filesystem without quota support is considered as disabled quota.

aschnell:
Avoiding redundant quota rescans is implemented by https://github.com/openSUSE/snapper/pull/633.

Issue #459, opened 2018-12-02
issues with building snapper with boost1.68


sww1235:
Hello: 

I am attempting to update void linux's package of snapper to 0.8.1 ( https://github.com/void-linux/void-packages/pull/5330 ) but am running into issues since we are on boost1.68. 

I am not experience enough to tell if the errors I am seeing in travis are snapper related or somehow boost related.

I would be glad to assist in any way, but got in over my head once things weren't building.

aschnell:
See https://github.com/void-linux/void-packages/pull/5330.

spencerjanssen:
NixOS has the same build failure since updating to btrfs-progs 4.19.  I tracked the issue down to kdave/btrfs-progs@c6039704c.  The `swap` macro defined in `kerncompat.h` is conflicting with Boost's use of `swap`.

sww1235:
@spencerjanssen thanks for finding that. I think your patch should fix it since we just need to get rid of the btrfs swap in this instance.

aschnell:
Should be "fixed/workarounded" by https://github.com/openSUSE/snapper/pull/460.

PR #460, opened 2018-12-03
Work around btrfs v4.19

spencerjanssen:
This is an attempt to address #459.  This is not the most elegant solution, but it happens to work for me.


aschnell:
Thanks for analysing the issue and making a patch.

PR #462, opened 2018-12-05
- fixed stringerror function

aschnell:
The test macro is provided like that in strerror(3) this way now. In older versions it was different.


shundhammer:
LGTM

aschnell:
Thanks.

PR #465, opened 2018-12-11
Fix regexp to detect 1-char volume group name.

kuriyama:
I got `Creating config failed (invalid filesystem type).` error when doing `create-config`.

It seems to be caused by my 1-char (`D`) volume group (that results in `/dev/mapper/D-thin` thin volume and `/dev/mapper/D-docker` provisioned volume).

My fix below assume volume group name will not start in `-` (and not include `-`).

### DIFF
--- a/snapper/Lvm.cc
+++ b/snapper/Lvm.cc
@@ -387,7 +387,7 @@ namespace snapper
     bool
     Lvm::detectThinVolumeNames(const MtabData& mtab_data)
     {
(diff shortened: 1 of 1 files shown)


aschnell:
The correct regex should be ```"/dev/mapper/(.*[^-])-([^-].*)$```. Some unit tests also seem appropriate.

aschnell:
See https://github.com/openSUSE/snapper/pull/552.

kuriyama:
Thanks!

PR #470, opened 2019-01-09
Fix data corruption leading to kernel panic.

ddawson:
I found that trying to use Snapper to create a Btrfs snapshot with a qgroup set causes the system to lock up, sometimes with a kernel panic message on the console. I tracked down the problem to this code. As it currently stands, `buffer` goes out of scope before the `ioctl` call can use it, and in my case, at least, it seems some of the data is already overwritten. Moving it out of the `if` statement lets this work properly.


aschnell:
Thanks for analysing the problem and providing a patch.

PR #472, opened 2019-01-22
- adapted to libbtrfs changes to resurrect special btrfs directory comparison (bsc#1111414)

aschnell:
For https://trello.com/c/LzTx1c2d/447-tw-1111414-p2-snappers-background-comparison-makes-the-system-unusable, https://bugzilla.suse.com/show_bug.cgi?id=1111414.


shundhammer:
Since this is basically used as a flag, it might also have been a boolean variable, but like this it does the job just as well.

LGTM.

aschnell:
Indeed, a bool would be fine. The patch is based on https://github.com/openSUSE/snapper/pull/438 (which has merge conflicts now) and btrfs-progs and there it is that way for whatever reasons.

PR #475, opened 2019-02-19
Revert "- removed unused variable" to fix build on Fedora

Conan-Kudo:
The variable previously removed is actually used when SELinux is enabled.

This reverts commit 842721cd470ce5b97316fff250b0c1fa356a00f4.


aschnell:
Thanks for the fix.

PR #478, opened 2019-03-22
Update de.po

clavinet:


aschnell:
Translations are done in Weblate, see https://github.com/openSUSE/snapper/blob/master/po/README.

PR #479, opened 2019-03-26
- fixed seg.fault during rollback if the previous default subvolume is missing in the snapshot list (bsc#1130273)

aschnell:
For https://trello.com/c/XgrLM80B/877-42-bug-1130273-snapper-btrfs-rollback-resulting-in-segmentation-fault or https://bugzilla.suse.com/show_bug.cgi?id=1130273.

The important piece is in clients/snapper.cc where so far the ```ProxySnapshots::iterator it = snapshots.find(previous_default.second)``` could result in the end iterator. This is now avoided by using proper iterators right away.


client/snapper.cc
```diff
@@ -1318,12 +1318,12 @@ command_rollback(ProxySnappers* snappers, ProxySnapper* snapper)
...
 
-    pair<bool, unsigned int> previous_default = filesystem->getDefault();
+    ProxySnapshots& snapshots = snapper->getSnapshots();
 
-    if (previous_default.first && scd1.description == "rollback backup")
-	scd1.description += sformat(" of #%d", previous_default.second);
+    ProxySnapshots::iterator previous_default = snapshots.getDefault();
 
-    ProxySnapshots& snapshots = snapper->getSnapshots();
+    if (previous_default != snapshots.end() && scd1.description == "rollback backup")
```
lslezak: I'd just prefer some `#define` or a constant to have a single place for this string. Currently you need to update two places when changing it and that's quite fragile...

PR #484, opened 2019-06-12
backport of extended cleanup algorithm

aschnell:
For https://trello.com/c/c2yIqaNH/980-snapper-backport-feature-enhanced-snapper-space-aware-clean-up-algorithm-to-ensure-minimal-free-space-is-available-to-sle-12-sp5.


aschnell:
Travis fails for openSUSE Leap due to the original header include problem of btrfsprogs. Likely the Docker image is outdated since that problem does not show up in the SUSE IBS.

The reason why Travis fails for Ubuntu is unclear to me. But it fails already during setup and seems unrelated to the PR.

In today's standup we agreed that these failures can be ignored for the time being.

shundhammer:
LGTM

jsrain:
@aschnell I believe that failures of SLE12-SPx packages should be ignored on openSUSE Leap 15.0 or anything newer not for the time being, but forever. Maybe I just got something wrong in the stand-up...

Issue #236, opened 2016-04-07
don't throw in create_subvolume() if subvolume exists


lnussel:
in order to be able to use mksubvolume in scripts, create_subvolume() should not throw an error if a subvolume already exists

laenion:
I was running into this problem when trying to use mksubvolume for a package script, so I decided to implement this enhancement.

PR #488, opened 2019-06-12
Add option to reuse existing subvolume

laenion:
If a subvolume exists already mksubvolume will abort. In certain
situations however, e.g. if there was a rollback to a previous snapshot,
it may be desirable to be able to reuse this subvolume.

This will make it possible to use the mksubvolume command in package
scripts.

To avoid having to duplicate the exception handler code multiple times
a C wrapper object was introduced to automatically clean up in the
destructor.

Fixes https://github.com/openSUSE/snapper/issues/236.

### DIFF
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -390,7 +390,6 @@ doit()
     mnt_table_parse_mtab(mtab_table, NULL);
 
     libmnt_fs* expected_fs = create_fstab_line(fs, subvol_option, subvolume_name);
-    char* expected_target = mnt_fs_get_target(expected_fs);
 
     // Consistency checks on (partially) existing entries
     libmnt_fs* fstab_entry = mnt_table.find_target(target, MNT_ITER_FORWARD);
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -322,7 +322,13 @@ doit()
 	throw runtime_error("invalid target");
 
     if (access(target.c_str(), F_OK) == 0)
-	throw runtime_error("target exists");
+    {
+	struct stat sb;
+	if (force && lstat(target.c_str(), &sb) == 0 && sb.st_mode & S_IFDIR)
+	    cout << "reusing target dir" << endl;
+	else
+	    throw runtime_error("target exists");
+    }
 
     if (access(dirname(target).c_str(), F_OK) != 0)
 	throw runtime_error("parent of target does not exist");
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -335,6 +335,8 @@ do_consistency_checks(MntTable& mnt_table, libmnt_fs* fs, libmnt_fs* expected_fs
 		throw runtime_error("subvolume of mounted target doesn't match");
 	}
     }
+
+    mnt_unref_cache(cache);
 }
 
 
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -186,13 +186,10 @@ do_add_fstab_and_mount(MntTable& mnt_table, const libmnt_fs* fs, const string& s
 
 
 void
-do_set_nocow()
+do_set_cow_flag()
 {
-    if (!set_nocow)
-	return;
-
     if (verbose)
-	cout << "do-set-nocow" << endl;
+	cout << "do-set-cow-flag" << endl;
 
     int fd = open(target.c_str(), O_RDONLY);
     if (fd == -1)
@@ -208,7 +205,10 @@ do_set_nocow()
 	throw runtime_error_with_errno("ioctl(EXT2_IOC_GETFLAGS) failed", errno);
     }
 
-    flags |= FS_NOCOW_FL;
+    if (set_nocow)
+	flags |= FS_NOCOW_FL;
+    else
+	flags &= ~FS_NOCOW_FL;
 
     if (ioctl(fd, EXT2_IOC_SETFLAGS, &flags) == -1)
     {
@@ -413,7 +413,7 @@ doit()
 
     do_add_fstab_and_mount(mnt_table, fs, subvol_option, subvolume_name);
 
-    do_set_nocow();
+    do_set_cow_flag();
 }
 
 
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -270,7 +270,7 @@ find_filesystem(MntTable& mnt_table)
 struct TmpMountpoint {
     unique_ptr<char[]> mountpoint;
     const libmnt_fs* fs;
-    TmpMountpoint(const string tmpMountpoint, const libmnt_fs* libmntfs, const string& subvol_opts)
+    TmpMountpoint(const string& tmpMountpoint, const libmnt_fs* libmntfs, const string& subvol_opts)
 	: mountpoint(strdup(tmpMountpoint.c_str())), fs(libmntfs)
     {
 	if (!mkdtemp(mountpoint.get()))
@@ -376,8 +376,8 @@ doit()
     }
     catch (const runtime_error_with_errno& e)
     {
-	if (e.error_number == EEXIST)
-	    cout << "subvolume exists already" << endl;
+	if (e.error_number == EEXIST && force)
+	    cout << "subvolume exists already, reusing" << endl;
 	else
 	    throw;
     }
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -140,21 +140,6 @@ do_add_fstab_and_mount(MntTable& mnt_table, const libmnt_fs* fs, const string& s
     if (!x)
 	throw runtime_error("mnt_copy_fs failed");
 
-    if (force)
-    {
-	struct libmnt_context* cxt = mnt_new_context();
-	libmnt_fs* y;
-	mnt_context_set_fs(cxt, x);
-	mnt_context_set_target(cxt, target.c_str());
-	if (!mnt_context_find_umount_fs(cxt, target.c_str(), &y))
-	{
-	    cout << "target mounted - umounting" << endl;
-	    int ret = mnt_context_umount(cxt);
-	    if (ret != 0)
-		throw runtime_error(sformat("mnt_context_umount failed, ret:%d", errno));
-	}
-    }
-
     mnt_fs_set_target(x, target.c_str());
 
     string full_subvol_option = subvolume_name;
@@ -168,20 +153,31 @@ do_add_fstab_and_mount(MntTable& mnt_table, const libmnt_fs* fs, const string& s
     mnt_fs_set_options(x, options);
     free(options);
 
-    // Caution: mnt_context_mount may change the source of x so the fstab
-    // functions must be called first.
-    mnt_table.add_fs(x);
-    mnt_table.replace_file();
+    if (mnt_table.find_target(target.c_str(), MNT_ITER_FORWARD) == NULL)
+    {
+	// Caution: mnt_context_mount may change the source of x so the fstab
+	// functions must be called first.
+	mnt_table.add_fs(x);
+	mnt_table.replace_file();
+    }
+    else
+	cout << "fstab entry exists already, skipping" << endl;
 
     if (mkdir(target.c_str(), 0777) != 0 && errno != EEXIST)
 	throw runtime_error_with_errno("mkdir failed", errno);
 
     struct libmnt_context* cxt = mnt_new_context();
-    mnt_context_set_fs(cxt, x);
+    libmnt_fs* y;
+    if (mnt_context_find_umount_fs(cxt, target.c_str(), &y))
+    {
+	mnt_context_set_fs(cxt, x);
 
-    int ret = mnt_context_mount(cxt);
-    if (ret != 0)
-	throw runtime_error(sformat("mnt_context_mount failed, ret:%d", ret));
+	int ret = mnt_context_mount(cxt);
+	if (ret != 0)
+	    throw runtime_error(sformat("mnt_context_mount failed, ret:%d", ret));
+    }
+    else
+	cout << "target mounted already, skipping" << endl;
 
     mnt_free_context(cxt);
 
@@ -266,20 +262,8 @@ find_filesystem(MntTable& mnt_table)
 	if (fs_fstype != "btrfs")
 	    throw runtime_error("filesystem is not btrfs");
 
-	if (fs_target == target)
-	{
-	    if (force)
-	    {
-		libmnt_fs* fs = mnt_table.find_target(target.c_str(), MNT_ITER_FORWARD);
-	        if (fs != NULL)
-		{
-		    cout << "removing existing fstab entry" << endl;
-		    mnt_table.remove_fs(fs);
-		}
-	    }
-	    else
-		throw runtime_error("target exists in fstab");
-	}
+	if (fs_target == target && !force)
+	    throw runtime_error("target exists in fstab");
 
 	if (!is_subvol_mount(fs_options))
 	    return fs;
--- a/client/mksubvolume.cc
+++ b/client/mksubvolume.cc
@@ -129,13 +129,10 @@ do_tmp_umount(const libmnt_fs* fs, const char* tmp_mountpoint)
 }
 
 
-void
-do_add_fstab_and_mount(MntTable& mnt_table, const libmnt_fs* fs, const string& subvol_option,
-		       const string& subvolume_name)
+libmnt_fs*
+create_fstab_line(const libmnt_fs* fs, const string& subvol_option,
+		  const string& subvolume_name)
 {
-    if (verbose)
-	cout << "do-add-fstab-and-mount" << endl;
-
     libmnt_fs* x = mnt_copy_fs(NULL, fs);
     if (!x)
 	throw runtime_error("mnt_copy_fs failed");
@@ -152,6 +149,16 @@ do_add_fstab_and_mount(MntTable& mnt_table, const libmnt_fs* fs, const string& s
     mnt_optstr_set_option(&options, "subvol", full_subvol_option.c_str());
     mnt_fs_set_options(x, options);
     free(options);
+
+    return x;
+}
+
+
+void
+do_add_fstab_and_mount(MntTable& mnt_table, libmnt_fs* x)
+{
+    if (verbose)
+	cout << "do-add-fstab-and-mount" << endl;
 
     if (mnt_table.find_target(target.c_str(), MNT_ITER_FORWARD) == NULL)
     {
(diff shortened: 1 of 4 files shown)


client/mksubvolume.cc
```diff
@@ -258,6 +267,33 @@ find_filesystem(MntTable& mnt_table)
     }
 }
 
+struct TmpMountpoint {
+    unique_ptr<char[]> mountpoint;
+    const libmnt_fs* fs;
+    TmpMountpoint(const string tmpMountpoint, const libmnt_fs* libmntfs, const string& subvol_opts)
```
aschnell: tmpMountpoint should be passed by reference.

client/mksubvolume.cc
```diff
@@ -332,36 +368,20 @@ doit()
...
-	    rmdir(tmp_mountpoint);
-	    free(tmp_mountpoint);
-	}
-	catch (...)
-	{
-	}
-
-	throw;
+	if (e.error_number == EEXIST)
+	    cout << "subvolume exists already" << endl;
```
aschnell: What happens if e.g. a directory already exists where the subvolume should be created?

client/mksubvolume.cc
```diff
@@ -43,6 +43,7 @@ using namespace BtrfsUtils;
 
 string target;
 
+bool force = false;
```
aschnell: Where is force used?

client/mksubvolume.cc
laenion: Not much, as nothing has to be done here - the destructor of TmpMountpoint will be called to clean up everything, and mksubvolume will continue with creating the fstab entry.

client/mksubvolume.cc
laenion: Oh, I'm sorry, this should have been used as part of the if condition in the exception handler above - will fix.

laenion: Thinking about it: It _could_ also just be left out, so existing suvolumes are reused automatically. I'm not sure if this is a good idea though.

client/mksubvolume.cc
aschnell: But adding the fstab entry for anything not a subvolume is wrong. This will lead to boot failures.

laenion: Ah, I see: You mean when it's just a directory instead of a subvolume. At least on openSUSE I hope that will never be the case, as that would mean someone modified the '@' subvolume, but I agree this may happen more easily on other distributions...

laenion: Detecting conflicting files in the latest commit now.

client/mksubvolume.cc
laenion: Done.

laenion:
Fixed all remarks. Checking for existing fstab entries is done in the code already, no changes were needed here.

client/mksubvolume.cc
```diff
@@ -377,7 +377,14 @@ doit()
     catch (const runtime_error_with_errno& e)
     {
 	if (e.error_number == EEXIST && force)
+	{
+	    const string path = string(tmp_mountpoint.mountpoint.get()).append("/").append(subvolume_name);
```
aschnell: string does have the + operator.

client/mksubvolume.cc
```diff
@@ -377,7 +377,14 @@ doit()
...
+
+	    struct stat sb;
+	    if (stat(path.c_str(), &sb) == 0 && !is_subvolume(sb))
```
aschnell: Failure of stat is not handled.

laenion: And it should be lstat anyway...

client/mksubvolume.cc
laenion: Changed.

client/mksubvolume.cc
laenion: Error condition reversed to only accept "good" case.

aschnell:
Isn't the operation stopped anyway due to the ``` if (access(target.c_str(), F_OK) == 0)```?

laenion:
Only if the target mount directory (in contrast to the subvolume itself) exists already. But looking at the rollback case the creation of this directory would also have been rolled back.
However: When initially implementing this patch I indeed had a check in that `if` query to not throw an error when the `-f` option was set. I can re-add that check if desired.

aschnell:
AFAIS the target mount directory can be identical to the subvolume if the "root" subvolume is mounted. That was the case with SUSE some time ago.

And rollback could not be the only usecase for the option. So I suggest to disable also that check.

aschnell:
If the entry already exists in /etc/fstab the operation also stops.

aschnell:
Maybe we could define that the following is a valid use-case: An RPM creates a subvolume in post-install but does not remove the subvolume in post-uninstall. Removing the RPM and installing it again must work.

laenion:
Changed the handling according to the (imho reasonable) use-case: An existing target directory will be reused now, and if the target exists in /etc/fstab the mount entry will be removed. If the target it mounted already an attempt to unmount it will be done, but if that fails the operation will still abort.

The new behavior will *only* be triggered with the `-f` option, though RPM `post` scripts should probably always call `mksubvolume` with this option for rollback compatibility...

aschnell:
What is the reason for first deleting the fstab entry and unmounting the subvolume and then add and mount again? Why not simply omit the add/mount step in that case?

laenion:
My thinking was that there may be artifacts from a previous run which may not match the new entry, but you are correct, this won't happen when not interfering manually. I'll change the code accordingly.

aschnell:
There might be other artifacts, e.g. options for the subvolume that are not changed (no-cow is not cleared).

aschnell:
What should happen if the user has already mounted a different btrfs filesystem at the target location?
That looks like a valid use-case to me.

Right now the program fails with an exception.

laenion:
After some thinking I tend to implement it to just create the missing parts, but with added consistency checks:
* Just leave out the option completely (currently `-f`) and behave reasonably well by default:
  * If the subvolume exists already check if the COW flag matches the command line option
  * If an fstab entry exists already check whether device and subvolume or subvolumeid match the expected new entry
  * If the target is mounted already also check for correct device and subvolume/-id
* If any of those conditions are not true abort with an error.

Advantages:
* Works without any modification to existing packaging scripts
* Exits with an error if the user changed something manually (i.e. the environment doesn't match the expected values which would have been autocreated), creates the missing entries otherwise
* No data loss by forcefully overwriting things
* No inconsistent data by just leaving entries as they are without consistency check

laenion:
The latest commits are implementing https://github.com/openSUSE/snapper/pull/488#issuecomment-511906588.

client/mksubvolume.cc
```diff
@@ -326,45 +378,79 @@ doit()
...
+	throw runtime_error("target is a dedicated mountpoint");
+    if (fstab_entry != NULL && strcmp(mnt_fs_get_source(fstab_entry), mnt_fs_get_source(expected_fs)) != 0)
+	throw runtime_error("existing fstab entry doesn't match target device");
+    if (fstab_entry != NULL && strcmp(mnt_fs_get_options(fstab_entry), mnt_fs_get_options(expected_fs)) != 0)
+	throw runtime_error("existing fstab entry options don't match");
+    // Something is mounted there already. Is it the correct device?
+    if (mounted_entry != NULL)
+    {
+        char* subvol_expected;
+	size_t valsz;
```
aschnell: AFAIS since valsz is not really needed passing NULL to mnt_fs_get_option should be fine.

client/mksubvolume.cc
```diff
@@ -326,45 +378,79 @@ doit()
...
+							mnt_fs_get_source(expected_fs), MNT_ITER_BACKWARD));
+	// Find last device in mtab to get the actual mount
+	mounted_entry = mnt_table_find_target(mtab_table, mnt_fs_get_target(expected_fs),
+					      MNT_ITER_BACKWARD);
+	if (mounted_entry != NULL)
+	{
+	    if (strcmp(real_device, mnt_fs_get_source(mounted_entry)) != 0)
+		throw runtime_error("different device mounted on target");
+	    char* subvol_real;
+	    mnt_fs_get_option(mounted_entry, "subvol", &subvol_real, &valsz);
```
aschnell: Is it OK to not check the return value? What it the subvol entry does not get found?

client/mksubvolume.cc
```diff
@@ -258,6 +277,33 @@ find_filesystem(MntTable& mnt_table)
     }
 }
 
+struct TmpMountpoint {
```
aschnell: I suggest to make mountpoint and fs private and add a get_path() function.

client/mksubvolume.cc
```diff
@@ -326,45 +378,79 @@ doit()
...
+
+    // Consistency checks on (partially) existing entries
+    libmnt_fs* fstab_entry = mnt_table.find_target(target, MNT_ITER_FORWARD);
+    libmnt_fs* mounted_entry = mnt_table_find_target(mtab_table, mnt_fs_get_target(expected_fs),
+						     MNT_ITER_BACKWARD);
+    if (subvolume_name.empty())
+	throw runtime_error("target is a dedicated mountpoint");
+    if (fstab_entry != NULL && strcmp(mnt_fs_get_source(fstab_entry), mnt_fs_get_source(expected_fs)) != 0)
+	throw runtime_error("existing fstab entry doesn't match target device");
+    if (fstab_entry != NULL && strcmp(mnt_fs_get_options(fstab_entry), mnt_fs_get_options(expected_fs)) != 0)
```
aschnell: Even harmless options like "rw" will make this check fail. Are there not options the user might have added that should not stop the program? Please at least add a comment.

client/mksubvolume.cc
laenion: Correct, I even did that in the reimplementation of `is_subvol_mount`, but didn't change this occurrence.
Will change.

client/mksubvolume.cc
```diff
@@ -326,45 +378,79 @@ doit()
...
+
+    libmnt_fs* expected_fs = create_fstab_line(fs, subvol_option, subvolume_name);
+
+    // Consistency checks on (partially) existing entries
+    libmnt_fs* fstab_entry = mnt_table.find_target(target, MNT_ITER_FORWARD);
+    libmnt_fs* mounted_entry = mnt_table_find_target(mtab_table, mnt_fs_get_target(expected_fs),
+						     MNT_ITER_BACKWARD);
+    if (subvolume_name.empty())
+	throw runtime_error("target is a dedicated mountpoint");
+    if (fstab_entry != NULL && strcmp(mnt_fs_get_source(fstab_entry), mnt_fs_get_source(expected_fs)) != 0)
```
aschnell: This also fails if the filesystem is mounted e.g. by label in fstab (and / is mounted by something else).

client/mksubvolume.cc
```diff
@@ -326,45 +378,79 @@ doit()
...
+    if (fstab_entry != NULL && strcmp(mnt_fs_get_options(fstab_entry), mnt_fs_get_options(expected_fs)) != 0)
+	throw runtime_error("existing fstab entry options don't match");
+    // Something is mounted there already. Is it the correct device?
+    if (mounted_entry != NULL)
+    {
+        char* subvol_expected;
+	size_t valsz;
+	mnt_fs_get_option(expected_fs, "subvol", &subvol_expected, &valsz);
+	// Map UUID / LABEL to a physical device name
+	const char* real_device = mnt_fs_get_source(mnt_table_find_source(mtab_table,
```
aschnell: real_device can be null when the entry for / is wrong in fstab (or udev did not generate the by-label link) in which case the program can core dump.

client/mksubvolume.cc
laenion: This was written under the premise that the application will usually be called by scripts with idempotent results from consecutive runs. Manually created or modified entries are deliberately not recognized as "matching", as the result is not the same.

This could be changed, but would significantly increase code complexity (similar to the mounted device detection). Shall I still implement it?

aschnell: Documenting it is also fine for me.

laenion: I've changed it to check only the `subvol` option (so any other option the user may add will be accepted).

As `subvol` may be given either relative or absolute (`@/something` vs. `/@/something`) a normalization will be done before the comparison.

client/mksubvolume.cc
laenion: This should not have happened, as we knew at this place that the device contains a btrfs file system, which will always have a `subvol` option. Just to be safe I've added the check, though.

client/mksubvolume.cc
laenion: Done.

client/mksubvolume.cc
laenion: The problem is that both `fstab_entry` and `expected_fs` are "virtual" entries without any guarantee that they are actually mounted. `real_device` however is determined by looking for the device in the _mtab_ file, because that one will always contain the absolute device path.

I'm not very keen on parsing everything manually, so for the time being I'd opt to not implement this.

client/mksubvolume.cc
laenion: Fixed.

client/mksubvolume.cc
```diff
@@ -326,45 +396,84 @@ doit()
...
-	{
-	    do_tmp_umount(fs, tmp_mountpoint);
-	    rmdir(tmp_mountpoint);
-	    free(tmp_mountpoint);
-	}
-	catch (...)
+	if (e.error_number == EEXIST)
 	{
-	}
+	    const string path = string(tmp_mountpoint.get_path()) + "/" + subvolume_name;
```
aschnell: The string constructor is not needed.

client/mksubvolume.cc
```diff
@@ -259,6 +278,51 @@ find_filesystem(MntTable& mnt_table)
...
+    }
+
+    ~TmpMountpoint()
+    {
+	do_tmp_umount(fs, mountpoint.get());
+	rmdir(mountpoint.get());
+    }
+
+    string
+    get_path()
```
aschnell: Should be a const function.

client/mksubvolume.cc
```diff
@@ -326,45 +396,84 @@ doit()
...
+    // Set up cache for UUID / LABEL resolution in mnt_table_find_source
+    libmnt_cache* cache = mnt_new_cache();
+    libmnt_table* mtab_table = mnt_new_table();
+    mnt_table_set_cache(mtab_table, cache);
+    mnt_table_parse_mtab(mtab_table, NULL);
+
+    char* subvol_expected;
+    libmnt_fs* expected_fs = create_fstab_line(fs, subvol_option, subvolume_name);
+    mnt_fs_get_option(expected_fs, "subvol", &subvol_expected, NULL);
+    assert(subvol_expected != NULL);
```
aschnell: Please throw an exception.

client/mksubvolume.cc
```diff
@@ -326,45 +396,84 @@ doit()
...
+
+    // Consistency checks on (partially) existing entries
+    libmnt_fs* fstab_entry = mnt_table.find_target(target, MNT_ITER_FORWARD);
```
aschnell: Below are two tests for fstab_entry != NULL. Why not throw an exception right here in that case?

aschnell:
Have all those pre checks been checked with multiple devices btrfs?

laenion:
> Have all those pre checks been checked with multiple devices btrfs?

No, I currently don't have such a setup. I won't make it to set it up before my vacation next week - could I ask you to run these tests?

aschnell:
Mutliple devices btrfs does not work correctly. I have a btrfs on sda2 and sda3. In fstab sda3 is used. Anyway in /proc/mounts sda2 is reported. Running the program a second time reports an error ("failure (different device mounted on target)").

BTW: Some debug output when using --verbose could help testing, e.g. the device names detected in this case.

laenion:
So in the end it seems a reliable consistency check is not possible, at least I'm not aware of any existing mechanism to get a canonical path for any given source identifier (even unmounted ones)...

I'm really tempted to just go back to the original use case and just ignore an already existing subvolume, without any additional checks or fixes...

laenion:
Correction: This may be what `mnt_resolve_spec` does.

client/mksubvolume.cc
laenion: Fixed.

client/mksubvolume.cc
laenion: Fixed.

client/mksubvolume.cc
laenion: Fixed by reworking error handling.

client/mksubvolume.cc
laenion: Fixed by using `mnt_resolve_spec`, which does exactly what we want.

laenion:
`mnt_resolve_spec` does indeed what I was looking for all the time - it will resolve any source to a canonical path spec. To ensure that the cache - required for this conversion to work - is always initialized I've added the option to set the cache to the MntTable class and changed the code to use it for all mnt_table operations.

client/mksubvolume.cc
```diff
@@ -144,20 +149,41 @@ do_add_fstab_and_mount(MntTable& mnt_table, const libmnt_fs* fs, const string& s
...
-    mnt_table.replace_file();
+    return x;
+}
+
+
+void
+do_add_fstab_and_mount(MntTable& mnt_table, libmnt_fs* x)
+{
+    if (verbose)
+	cout << "do-add-fstab-and-mount" << endl;
```
lslezak: I know this line was already there, but what's the purpose of this? Is it a debugging message or something for the end user? For debugging it should use `y2deb`/`y2mil`, for end users it should contain a better text and should be translated...

laenion:
It seems `mksubvolume` is not using any of the _y2_ helper functions. In the case of the output example from above it seems the messages are just there for debugging purposes / to know where the application currently is.
However: Switching the message mechanism of the applications to a more reasonable one is imho out of the scope for this patch.

For this patch I just used the existing mechanism for messages, but I could hide the additional output I introduced behind the `verbose` switch (just like the existing code does, though I would definitely prefer to print the messages in any case to be able to see what happened during execution in the log files).
Please let me know how we should proceed.

client/mksubvolume.cc
```diff
@@ -326,45 +396,83 @@ doit()
...
+    if (mnt_fs_get_option(expected_fs, "subvol", &subvol_expected, NULL) != 0)
+	throw runtime_error("mnt_fs_get_option failed");
+
+    // Consistency checks on (partially) existing entries
+    libmnt_fs* fstab_entry = mnt_table.find_target(target, MNT_ITER_FORWARD);
+    libmnt_fs* mounted_entry = mtab_table.find_target(mnt_fs_get_target(expected_fs),
+						      MNT_ITER_BACKWARD);
+    if (subvolume_name.empty())
+	throw runtime_error("target is a dedicated mountpoint");
+    if (fstab_entry != NULL && strcmp(mnt_resolve_spec(mnt_fs_get_source(fstab_entry), cache),
```
aschnell: At least mnt_resolve_spec () can return NULL causing a seg.fault. I can trigger this by writing a wrong UUID for / in fstab.

laenion: Fixed, thanks.

client/mksubvolume.cc
```diff
@@ -326,45 +396,85 @@ doit()
...
 
-    string subvolume_name = target.substr(fs_target.size() + (fs_target == "/" ? 0 : 1));
+    string subvolume_name = target.substr(fs_target.size() +
+					 (fs_target == "/" || fs_target == target ? 0 : 1));
     if (verbose)
 	cout << "subvolume-name:" << subvolume_name << endl;
 
-    // Execute all steps.
+    // Set up cache for UUID / LABEL resolution in mnt_table_find_source
+    libmnt_cache* cache = mnt_new_cache();
```
aschnell: mnt_new_cache() could fail. More important is when cache gets freed, AFAIS a call to mnt_unref_cache() somethere later is missing.

aschnell:
The function doit() increased a lot. Moving some code to helper functions could help a lot to understand the code.

client/mksubvolume.cc
laenion: Error handling for `mnt_new_cache` is handled in `MntTable::set_cache` already. MntTable will now also dereference the cache automatically.

laenion:
The consistency check code has been moved to a separate function now.

aschnell:
Unfortunately I have found another problem with the checks. A multiple device btrfs mounted by the device name in /etc/fstab. In /proc/mounts another device of the btrfs can be used and then the
check fails and reports "different device mounted on target".

client/mksubvolume.cc
```diff
@@ -259,6 +278,104 @@ find_filesystem(MntTable& mnt_table)
...
+    {
+	if (mnt_fs_get_option(fstab_entry, "subvol", &subvol_fstab, NULL) != 0 ||
+		get_abs_subvol_path(subvol_fstab) != get_abs_subvol_path(subvol_expected))
+	    throw runtime_error("existing fstab entry's subvolume doesn't match");
+    }
+    // Something is mounted there already. Is it the correct device?
+    if (mounted_entry != NULL)
+    {
+	// Find last device in mtab to get the actual mount
+	mounted_entry = mtab_table.find_target(mnt_fs_get_target(expected_fs), MNT_ITER_BACKWARD);
```
aschnell: Isn't the exact statement already used 20 lines above to set ```mount_entry```?

client/mksubvolume.cc
```diff
@@ -259,6 +278,104 @@ find_filesystem(MntTable& mnt_table)
...
+	subvolume.insert(0, "/");
+    return subvolume;
+}
+
+
+void
+do_consistency_checks(MntTable& mnt_table, libmnt_fs* fs, libmnt_fs* expected_fs)
+{
+    char *subvol_expected, *subvol_fstab, *subvol_real;
+    const char *dev_expected, *dev_fstab;
```
aschnell: These variables should be declares as late as possible. Some are even only used in sub blocks.

client/mksubvolume.cc
laenion: Yes, it is - I forgot to remove that after several code refactorings...

laenion:
I'm not sure where the Changelog is stored, so here are the corresponding ticket numbers:

* https://bugzilla.suse.com/show_bug.cgi?id=1138725 (boo#1138725: Add Ignition to Tumbleweed)
* https://bugzilla.suse.com/show_bug.cgi?id=1126900 (boo#1126900: health-checker: grub2-editenv: error: cannot open `/boot/grub2/grubenv': Read-only file system.)
* https://github.com/openSUSE/snapper/issues/236 (don't throw in create_subvolume() if subvolume exists)

aschnell:
The changelog is in package/snapper.changes

aschnell:
looks good now

jpeaglesandkatz:
This problem is not fixed.

CachyOS latest updated 6.14 kernel.. KDE plasma 

" Creating config failed (creating btrfs subvolume .snapshots failed since it already exists)." 

I had to remove and reinstall snapper/btrfs suppor assistant because it was interfering with timeshift.

upon reinstall I cannot create a config in btrfs assistant because it complains the .btrfs subvolume already exists. Which seems obvious since they containt the snapshots from my previously installed snapper support.

How do I fix this issue now? I cannot use snapper/btrfs assistant at all.

PR #492, opened 2019-09-16
 Move D-Bus conf file to share/dbus-1/system.d 

worldofpeace:
Since D-Bus 1.9.18 configuration files installed by third-party should
go in share/dbus-1/system.d. The old location is for sysadmin overrides.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -146,6 +146,14 @@ if test "x$with_selinux" = "xyes"; then
 	AC_CHECK_LIB(selinux, selinux_snapperd_contexts_path, [], [AC_MSG_ERROR([selinux library does not provide selinux_snapperd_contexts_path symbol])])
 fi
 
+AC_ARG_WITH(dbus-sys, [  --with-dbus-sys=<dir>   where D-BUS system.d directory is])
+if ! test -z "$with_dbus_sys" ; then
+    DBUS_SYS_DIR="$with_dbus_sys"
+else
+    DBUS_SYS_DIR="$datadir/dbus-1/system.d"
+fi
+AC_SUBST(DBUS_SYS_DIR)
+
 PKG_CHECK_MODULES(DBUS, dbus-1)
 
 AC_CHECK_HEADER(acl/libacl.h,[],[AC_MSG_ERROR([Cannout find libacl headers. Please install libacl-devel])])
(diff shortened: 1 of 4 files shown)


aschnell:
Thanks for the contribution.

Since when does D-Bus also check in /usr/share? snapper is also provided for older distributins (see https://build.opensuse.org/package/show/filesystems:snapper/snapper) and those must not break.

worldofpeace:
Since 1.9.18 released in 2015. Which distro's do you deliver to that have DBus older than this?

aschnell:
In the build project I mentioned there are some, e.g. SLE12 SP4, RHEL 7, Fedora 22, Debian 8 or Ubuntu 15.04 all use dbus <= 1.8. At least SLE12 SP4 is still maintained by SUSE.

Is there no way to ask dbus about configuration settings at build time?

worldofpeace:
I will see if I can arrange something in `configure.ac` to use the old directory when it doesn't match that version.

worldofpeace:
How does this seem?
```
AC_ARG_WITH(dbus-sys, [  --with-dbus-sys=<dir>   where D-BUS system.d directory is])
if ! test -z "$with_dbus_sys" ; then
    DBUS_SYS_DIR="$with_dbus_sys"
else
    DBUS_SYS_DIR="$datadir/dbus-1/system.d"
fi
AC_SUBST(DBUS_SYS_DIR)
```

And in your specs you can use `--with-dbus-sys=/etc/dbus-1/system.d` for older targets?

aschnell:
Yes, as a last resort setting the path like that should work.

mvidner:
Hi @worldofpeace , I don't know about building on Debian but I managed to fix the build failure anyway :) https://github.com/openSUSE/snapper/commit/67d299b1036cb4c5658825d52cc49b436d131493

worldofpeace:
> Hi @worldofpeace , I don't know about building on Debian but I managed to fix the build failure anyway :) [67d299b](https://github.com/openSUSE/snapper/commit/67d299b1036cb4c5658825d52cc49b436d131493)

Thanks, will integrate :+1:

aschnell:
Meanwhile the distributions where dbus is too old are dropped for other reasons for snapper. So the conditional stuff is not needed more.

But at least on SLE 15 the directory /usr/share/dbus-1/system.d is not yet owned by systemd. I'm cheking all supported distributions right now.

aschnell:
Implemented with https://github.com/openSUSE/snapper/pull/634.

PR #496, opened 2019-10-14
Allow trailing comments in configs (bsc#1005342)

mvidner:
## Problem

- [bsc#1005342](https://bugzilla.suse.com/show_bug.cgi?id=1005342)
- https://trello.com/c/hCzKci5i

A user has reported that adding a trailing comment to a config line makes Snapper ignore that line:
```sh
TIMELINE_LIMIT_MONTHLY="0"      # No monthly snapshots
```

This is contrary to the expectation that these config files follow the shell syntax.

## Solution

I've fixed these methods of SysconfigFile to deal with trailing comments:

- [x] getValue
- [x] getAllValues
- [x] setValue

## TODO

- [x] changelogs, version


mvidner:
A look at the implementation shows there are still problems with quoting. For example this ends with an error, creating a config Snapper cannot parse:
```console
# snapper -c '5"1/4' create-config /local/btrfs
Failure (error.something).
# grep = /etc/sysconfig/snapper 
SNAPPER_CONFIGS="5"1/4"
```

.gitignore
```diff
@@ -17,4 +17,5 @@ obs-package-from-git
 stamp-h1
 py-compile
 test-driver
+testsuite/*.txt.tmp
```
lslezak: NP: maybe a generic `testsuite/*.tmp` (or even global `*.tmp`) would be better (future proof)...

testsuite/sysconfig-get1.cc
```diff
@@ -42,4 +43,29 @@ BOOST_AUTO_TEST_CASE(sysconfig_get1)
...
+    s.setValue("K2", "all new");
+    BOOST_CHECK(s.getValue("K2", tmp_string));
+    BOOST_CHECK_EQUAL(tmp_string, "all new");
+
+    s.setValue("K2", "changeme");
+    BOOST_CHECK(s.getValue("K2", tmp_string));
+    BOOST_CHECK_EQUAL(tmp_string, "changeme");
+
+    s.save();
+    BOOST_CHECK_EQUAL(system("diff -u sysconfig-set1.txt sysconfig-set1.txt.tmp"), 0);
```
lslezak: NP: That `.tmp` file is marked in .gitignore, but maybe we could delete it if the test succeeds, I do not like leaving garbage on the disk...

aschnell: It should also be possible to make the comparison in memory.

README.md
```diff
@@ -79,23 +79,28 @@ package are mentioned in the `package/snapper.changes` file, update also the
...
+  OBS project has to be updated.
+
+- To create the package use command `make package`. Then use the common
+  work-flow to submit the
+package to the build service. For [openSUSE:Factory][]
+  send at first the package to the devel project [filesystems:snapper][] in OBS.
+
+    *Please note that this OBS project builds for more distributions
+    so more metadata files have to be updated. See the OBS documentation
+    for more info ([cross distribution howto][xdist], [Debian builds][]).*
```
mvidner: @aschnell BTW I have checked the build targets in https://build.opensuse.org/package/show/filesystems:snapper/snapper and it includes several xUbuntu versions that are long out of support. How about leaving only the LTS ones, 16.04, 18.04 + newer?

<details>
<summary>Ubuntu_version_history</summary>
https://en.wikipedia.org/wiki/Ubuntu_version_history
"The support period for non-LTS releases is 9 months."

Ubuntu 14.04 LTS (Trusty Tahr)
"Normal LTS support is set to continue until 25 April 2019, after which extended security maintenance will be available to Ubuntu Advantage customers and as a separate commercial purchase, as was the case previously with 12.04."

Ubuntu 16.04 LTS (Xenial Xerus)
Ubuntu 18.04 LTS (Bionic Beaver)

https://xubuntu.org/releases/ : "Regular Xubuntu releases are supported for 9 months and the LTS releases for 3 years."
</details>

Similarly it includes oS 13.1 which is 3 years out of support: https://en.opensuse.org/Lifetime

Fedora 20 is 4 years out of support: https://fedoraproject.org/wiki/End_of_life

aschnell: Since the build service still provides those versions it seems as if somebody sees a use case.
On the other hand, what is gained for snapper by removing them? Can snapper use newer language or library features to simplify code?

mvidner: Well, there *is* a 6 year old piece of conditional compilation that we could remove:
https://github.com/openSUSE/snapper/blob/ec58ba76a4f86405184b0ba4c6c2fa274a7c0469/snapper/Btrfs.cc#L1371-L1392

OTOH you're right, we should not fix what ain't broken. IMHO we should touch it only once something breaks. I just got curious seeing so many build targets.

aschnell: That code could indeed be removed then. When it comes to threads it might be worth to switch from boost to C++11. Same for regex.

testsuite/sysconfig-get1.cc
mvidner: I tried to do it without writing the temporary file, by using AsciiFile::lines and then BOOST_CHECK_EQUAL_COLLECTIONS, but SysconfigFile is a *protected* subclass of AsciiFile so that method is inaccessible. @aschnell did you have another way in mind?

aschnell: OK, lines() can indeed not be used right away. So some extension to SysConfigFile would be needed.

mvidner:
I keep the original scope, fixing only the trailing comments.

PR #501, opened 2019-10-23
Pythonless plugin

mvidner:
- https://trello.com/c/1yxNFV0R
- https://jira.suse.com/browse/SLE-10500

In order to be able to support some products for a looong time, we want to get rid of dependencies with a short support cycle.
For Snapper, that means removing the Python dependency for its [ZYpp plugin](https://doc.opensuse.org/projects/libzypp/SLE12SP2/zypp-plugins.html) (which makes snapshots whenever an "interesting" package is installed or updated).

### Implementation notes

- the Regular Expression dialect has changed; see the Release Notes section below
- the plugin tests in the end do not depend on the integration test framework introduced in PR #504, because they do not reach a live snapperd
- there's a new testing-only dependency on Ruby + ruby-dbus.gem, used to run a mock snapperd. It's used in Travis but skipped in OBS.
- `override` keywords are now used in C++ code, with `-Wsuggest-override` in Makefiles
- in snapper::XmlFile, the methods `getChildNode` and similar were fixed to actually operate on children, whereas previously they operated on siblings. So I bumped the library so-version because the fix keeps the ABI but changes the behavior. The same problem exists in libstorage-ng.

### Proposed Release Notes Entry

snapper-zypp-plugin, a plugin for ZYpp that makes filesystem snapshots with Snapper during commits, was rewritten (to remove a dependency on Python and thus enhance long-term supportability). As a result, Python regular expressions are no longer supported in the /etc/snapper/zypp.conf file. POSIX extended regular expressions work instead, which should work sufficiently well for the purpose of package name matching. Shell patterns continue working unchanged.

### TODO

- [x] version + changelog
- [x] actually parse the *userdata*
- [x] test XML (config) that is not well formed or invalid
- [x] test JSON (zypp data) that is not well formed or invalid
- [x] switch from rapidjson to [json-c](https://github.com/json-c/json-c)
- [x] switch from boost::regex to snapper::Regex
- [x] test and describe the switch of the regex dialect


zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,363 @@
+// getenv
+#include <stdlib.h>
+// getppid
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <boost/format.hpp>
```
aschnell: There is already sformat in storage/AppUtil.h. Is something else really needed (esp. something else that has not been included in C++20)?

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,296 @@
...
+
+map<string, string> SnapperZyppPlugin::get_userdata(const Message&) {
+    map<string, string> result;
+    return result;
+}
+
+set<string> SnapperZyppPlugin::get_solvables(const Message& msg, bool todo) {
+    set<string> result;
+
+    rapidjson::Document doc;
```
joseivanlopez: Umm, if we need a json library to import data, maybe we might modify the `JsonFormatter` implementation to use it.

mvidner: Oh, good point, I'll check that

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Thanks for the pointer, I will check that

zypp-plugin/testsuite/test1
```diff
@@ -0,0 +1,78 @@
...
+    $MYDIR/mock-snapperd &
+    sleep 1
+    PID=$!
+    trap "kill \$PID" EXIT TERM INT
+}
+teardown() {
+    :
+}
+
+# FIXME: run this automatically, and declare the ruby-dbus dependency
```
mvidner: Done

client/proxy-lib.h
```diff
@@ -162,7 +162,7 @@ class ProxySnappersLib : public ProxySnappers::Impl
 
     virtual map<string, ProxyConfig> getConfigs() const override;
 
-    virtual vector<string> debug() const { return Snapper::debug(); }
+    virtual vector<string> debug() const  override{ return Snapper::debug(); }
```
aschnell: Wrong spacing.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,349 @@
...
+            bus = DBUS_BUS_SESSION;
+        }
+    }
+};
+
+class SnapperZyppPlugin : public ZyppCommitPlugin {
+public:
+    struct SolvableMatcher {
+	enum class Kind {
+	    Glob,
```
aschnell: enum values are all uppercase in snapper.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,349 @@
...
+	    Regex
+	};
+	string pattern;
+	Kind kind;
+	bool important;
+
+	SolvableMatcher(string apattern, Kind akind, bool aimportant)
```
aschnell: apattern should be passed as const reference.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,349 @@
...
+	bool match(const string& solvable) {
+	    y2deb("match? " << solvable << " by " << ((kind == Kind::Glob)? "GLOB '": "REGEX '") << pattern << '\'');
+	    bool res;
+	    if (kind == Kind::Glob) {
+		static const int flags = 0;
+		res = fnmatch(pattern.c_str(), solvable.c_str(), flags) == 0;
+	    }
+	    else {
+		// POSIX Extended Regular Expression Syntax
+		boost::regex rx_pattern(pattern, boost::regex::extended);
```
aschnell: regex compilation can fail (which would be a user error). Some error logging seems appropriate.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,349 @@
...
+
+	static vector<SolvableMatcher> load_config(const string& cfg_filename) {
+	    vector<SolvableMatcher> result;
+
+	    XmlFile config(cfg_filename);
+	    // FIXME test parse errors
+	    const xmlNode* root = config.getRootElement();
+	    const xmlNode* solvables_n = getChildNode(root, "solvables");
+	    const list<const xmlNode*> solvables_l = getChildNodes(solvables_n, "solvable");
+	    for (auto it = solvables_l.begin(); it != solvables_l.end(); ++it) {
```
aschnell: Range based loop is allowed.

zypp-plugin/snapper_zypp_plugin.cc
aschnell: But maybe we should wait for Boost.JSON (https://github.com/vinniefalco/json). That should if accepted by available everywhere where boost is available (and that is required anyway).

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,349 @@
...
+    DBusBusType bus;
+
+    ProgramOptions()
+    : plugin_config("/etc/snapper/zypp-plugin.conf")
+    , snapper_config("root")
+    , bus(DBUS_BUS_SYSTEM)
+    {
+	const char * s;
+
+	s = getenv("SNAPPER_ZYPP_PLUGIN_CONFIG");
```
aschnell: The env. var. should be explained, esp. that they are only used for debugging AFAIS.

snapper/XmlFile.cc
```diff
@@ -122,23 +123,38 @@ namespace snapper
...
+    getValue(const xmlNode* node, string& value)
+    {
+	xmlChar* tmp = xmlNodeGetContent(node);
+	if (tmp == nullptr)
+	    value = "";
+	else {
+	    value = (const char *) tmp;
+	    xmlFree(tmp);
+	}
+	return true;
```
aschnell: This always return true. Please explain.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,349 @@
...
+	    }
+	}
+    }
+}
+
+bool
+log_query(LogLevel level, const string& component)
+{
+    if (level == DEBUG)
+	return getenv("DEBUG") != nullptr;
```
aschnell: That env. var. looks too generic. Or is that intentional?

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Example from zypp-plugin/testsuite/3-invalid-xml.test,
This will appear in zypper.log:
```console
DEBUG:match? mypkg by REGEX 'an invalid [ regex'
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::regex_error> >'
  what():  Unmatched [ or [^ in character class declaration.  The error occurred while parsing the regular expression fragment: 'id [ regex>>>HERE>>>'.
```

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Does the new comment of class ProgramOptions look good?

snapper/XmlFile.cc
mvidner: You're right, that was inconsistent. Now returning false if there's an error

Makefile.ci
```diff
@@ -14,7 +14,7 @@ package: obs-package-from-git
 # A script that is shared with other projects and should be always refreshed
 URL=https://raw.githubusercontent.com/openSUSE/obs-package-from-git/master/obs-package-from-git
 
-.PHONY: always_out_of_date
-
-obs-package-from-git: always_out_of_date
-	wget -O $@ $(URL) && chmod +x $@
+#.PHONY: always_out_of_date
```
mvidner: Oops, this must be reverted

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Now slightly changed after kicking out boost_regex:
```console
DEBUG:match? mypkg by REGEX 'an invalid [ regex'
terminate called after throwing an instance of 'std::runtime_error'
  what():  Regex compilation error: Unmatched [ or [^
```

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Currently I am linking to json-c where the *so* file has 64208 bytes. Some boost libraries have runtimes around a megabyte big, so we should check that factor too.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,341 @@
...
+	    catch (const Exception& ex) {
+		SN_CAUGHT(ex);
+	    }
+	}
+
+	return ack();
+    }
+
+    Message commit_end(const Message& msg) override {
+	cerr << "INFO:" << "COMMITEND" << endl;
```
joseivanlopez: NP: why `cerr` ? I would expect to use `clog` here.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,341 @@
...
+    set<string> get_solvables(const Message&, Phase phase);
+
+    void match_solvables(const set<string>& solvables, bool& found, bool& important);
+
+    unsigned int create_pre_snapshot(string config_name, string description, string cleanup, map<string, string> userdata);
+};
+
+const string SnapperZyppPlugin::cleanup_algorithm = "number";
+
+map<string, string> SnapperZyppPlugin::get_userdata(const Message& msg) {
```
joseivanlopez: NP: why are there some methods defined outside the class declaration but others inside?

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,341 @@
...
+    map<string, string> result;
+    auto it = msg.headers.find("userdata");
+    if (it != msg.headers.end()) {
+	const string& userdata_s = it->second;
+        vector<string> key_values;
```
joseivanlopez: NP: leftover indent?

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,341 @@
...
+                    continue;
+                }
+                if (json_object_get_type(name) != json_type_string) {
+                    cerr << "ERROR:" << "\"n\" is not a string" << endl;
+                    cerr << "ERROR:" << "in item #" << i << endl;
+                    continue;
+                }
+                else {
+                    const char * prize = json_object_get_string(name);
+		    result.insert(prize);
```
joseivanlopez: NP: looks like a missing indent.

zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -0,0 +1,341 @@
...
+		if (found && important)
+		    return; // short circuit
+	    }
+	}
+    }
+}
+
+int main() {
+    if (getenv("DISABLE_SNAPPER_ZYPP_PLUGIN") != nullptr) {
+	cerr << "INFO:" << "$DISABLE_SNAPPER_ZYPP_PLUGIN is set - disabling snapper-zypp-plugin" << endl;
```
joseivanlopez: NP: why `cerr`?

zypp-plugin/solvable_matcher.cc
```diff
@@ -0,0 +1,68 @@
...
+        << "match? " << solvable
+        << " by " << ((kind == Kind::GLOB)? "GLOB '": "REGEX '")
+        << pattern << '\'' << endl;
+    bool res;
+    if (kind == Kind::GLOB) {
+        static const int flags = 0;
+        res = fnmatch(pattern.c_str(), solvable.c_str(), flags) == 0;
+    }
+    else {
+        // POSIX Extended Regular Expression Syntax
```
joseivanlopez: NP: indentation.

joseivanlopez:
NP: In general I miss the license header in the new files. I guess we should add them.

data/zypp-plugin.conf
```diff
@@ -1,12 +1,21 @@
...
 
-  <!-- List of solvables. If the zypp commit transaction includes any solvable
-       listed below, snapper will create pre and post snapshots. The match
-       attribute defines whether the pattern is a Unix shell-style wildcard
-       ("w") or a Python regular expression ("re"). If any of the matching
-       solvables is marked as important, the snapshots are also marked as
-       important. -->
+       Shell-style wildcards, match="w"
+           Must cover the whole solvable name: the pattern "systemd*" matches
+           systemd-devel, but "systemd" does not.
```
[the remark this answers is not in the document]
mvidner: The glob `systemd` matches the package systemd but not systemd-devel.
The glob `systemd*` matches the packages systemd and systemd-devel.
The regex `systemd` matches the packages systemd and systemd-devel.
The regex `systemd$`  matches the package systemd but not systemd-devel.

zypp-plugin/zypp_plugin.h
```diff
@@ -0,0 +1,24 @@
+#include <map>
```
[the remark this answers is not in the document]
mvidner: My mistake, I will add them

zypp-plugin/testsuite/1-happy.test
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -e
+set -u
+#set -x
```
joseivanlopez: leftover?

zypp-plugin/snapper_zypp_plugin.cc
mvidner: 1. I forgot about [`clog`](http://www.cplusplus.com/reference/iostream/clog/). Duh.
2. Since the difference is that `clog` is a plain stream whereas `cerr` makes `cout` flush and also flushes itself, I think `cerr` is better for stdio-based interprocess communication: in case this plugin crashes, we want libzypp to have received the recent log messages.

zypp-plugin/solvable_matcher.cc
mvidner: yup, I need to regularize it

zypp-plugin/testsuite/tap-driver.sh
```diff
@@ -0,0 +1,651 @@
+#! /bin/sh
```
joseivanlopez: Isn't this script provided by any other package (e.g., automake)?

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Followup card: https://trello.com/c/L0GJlnET/3706-snapper-use-a-library-for-json-output

zypp-plugin/testsuite/tap-driver.sh
mvidner: Yes and no, good point, I'll improve the commit message to explain.

zypp-plugin/snapper_zypp_plugin.cc
mvidner: So I assume this is OK.

zypp-plugin/snapper_zypp_plugin.cc
mvidner: I assume it is good enough.

zypp-plugin/testsuite/tap-driver.sh
mvidner: Done: https://github.com/openSUSE/snapper/pull/501/commits/c804a7af8d56f1a84e90745ca1c032d285237418

mvidner: Done:  https://github.com/openSUSE/snapper/pull/501/commits/c804a7af8d56f1a84e90745ca1c032d285237418

zypp-plugin/snapper_zypp_plugin.cc
mvidner: Um, that is kind of random. Originally I put the "main logic" inline and the helpers outside.

mvidner:
> I have found minor things only. The test suite looks also ok, but take into account my limited bash knowledge. I would like somebody else have a look as well.

BTW the tests also test themselves for shell bugs, using https://www.shellcheck.net/ ;-)

mvidner:
@aschnell Arvin, can you please copy the tarball from https://build.opensuse.org/project/users/filesystems:snapper to the FTP site?

PR #506, opened 2019-11-24
Import some musl fixes from Alpine Linux

maxice8:
There are more patches to come but they need more time


mvidner:
For the curious, the Alpine Linux patches are here: https://github.com/alpinelinux/aports/tree/66a0a541d83acb73b22c9133a9c86175019b6f48/testing/snapper

Issue #12, opened 2013-04-09
Falha (org.freedesktop.DBus.Error.ServiceUnknown)


gnumoksha:
On debian 6:

aptitude install snapper
[...]
root@testes:/etc/snapper/configs# snapper list-configs
Falha (org.freedesktop.DBus.Error.ServiceUnknown).

Issue #10, opened 2013-03-02
rename configure.in to configure.ac


hasufell:
configure.in name is deprecated an will probably become fatal in automake-1.14

http://www.flameeyes.eu/autotools-mythbuster/forwardporting/automake.html

PR #510, opened 2019-12-06
add description to the second snapshot of rollback

hrotkogabor:
When we do a rollback, snapper will create two snapshots.
For example, the default subvolume is number 10, the latest is number 15, and the one we want to go back, is number 12.
Rollback will create number 16 from number 10, and number 17 from number 12 as the new writable default.
The first snapshot, number 16 will get the description "rollback backup of #10", but the other one will not get any description. This patch will modify it, to add the description to the second snapshot also, like this: "rollback backup of #12". This way it is easy to identify where the snapshot came from.

### DIFF
--- a/client/snapper.cc
+++ b/client/snapper.cc
@@ -1110,7 +1110,7 @@ command_rollback(cli::GlobalOptions* global_options, ProxySnappers* snappers, Pr
 	if (!global_options->quiet())
 	    cout << sformat(_("Creating read-write snapshot of snapshot %d."), tmp->getNum()) << flush;
 
-	if (tmp != snapshots.end() && scd2.description == default_description)
+	if (tmp != snapshots.end() && scd2.description == default_description2)
 	    scd2.description += sformat(" of #%d", tmp->getNum());
 
 	scd2.read_only = false;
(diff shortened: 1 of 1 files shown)


aschnell:
That looks wrong when a number is provided to the rollback command.

hrotkogabor:
Thank you pointing that out. I fixed that, and tested.

aschnell:
The numbers are fine now, but the snapshot is not really a backup since it is read-write and the new system. So a text like "new system of #" would be better. Maybe you can find an even better text.

hrotkogabor:
I changed the description to this: "writable copy of #".

client/snapper.cc
```diff
@@ -1104,6 +1110,9 @@ command_rollback(cli::GlobalOptions* global_options, ProxySnappers* snappers, Pr
 	if (!global_options->quiet())
 	    cout << sformat(_("Creating read-write snapshot of snapshot %d."), tmp->getNum()) << flush;
 
+	if (tmp != snapshots.end() && scd2.description == default_description)
```
aschnell: Has to be ```default_description2```.

Please rename ```default_description``` to ```default_description1```.

hrotkogabor: Sorry I missed that, corrected.

aschnell:
Thanks, looks fine now.

PR #514, opened 2019-12-19
musl build fix

ndowens:
This fixes builds for both glibc and musl

Signed-off-by: Nathan Owens <ndowens04@gmail.com>

### DIFF
--- a/snapper/AppUtil.cc
+++ b/snapper/AppUtil.cc
@@ -223,7 +223,7 @@ namespace snapper
     string
     stringerror(int errnum)
     {
-#if (_POSIX_C_SOURCE >= 200112L) && ! _GNU_SOURCE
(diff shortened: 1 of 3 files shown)


snapper/AppUtil.cc
```diff
@@ -223,7 +223,7 @@ namespace snapper
     string
     stringerror(int errnum)
     {
-#if (_POSIX_C_SOURCE >= 200112L) && ! _GNU_SOURCE
+#if (_POSIX_C_SOURCE >= 200112L) && ! _GNU_SOURCE || MUSL_LIBC
```
ericonr, line 226: This requires packagers to define `MUSL_LIBC` when building snapper, since musl doesn't define it itself.

PR #517, opened 2020-01-07
Fix several build problems

mvidner:
... that I created while implementing the previous release, duh.

https://bugzilla.suse.com/show_bug.cgi?id=1160306

1. It fails to build for Debian and Ubuntu in https://build.opensuse.org/package/show/filesystems:snapper/snapper (because the build dependencies need to be declared in other files than I expected)
2. On i586, /usr/lib64/snapper becomes /usr/lib/snapper so some files end up in 2 subpackages and they have file conflicts.
3. The C++ compiler flag -Wsuggest-override causes an error for older compilers (on SLE 12, Fedora 21)


aschnell:
LGTM

PR #520, opened 2020-01-16
Fix "Snapper is not creating the post snapshot" (bsc#1160938)

mvidner:
[bsc#1160938](https://bugzilla.suse.com/show_bug.cgi?id=1160938) is an embarrassing case of a forgotten debugging printf, which throws the communication protocol out of sync.

My new code validates the protocol that it receives, but not what it itself sends. I fix it by running the tests with another plugin working as a validating middleware (forwarding-zypp-plugin), which does catch this bug.

Also fixing [bsc#1160891](https://bugzilla.suse.com/show_bug.cgi?id=1160891) "make check fails for snapper after make clean".


zypp-plugin/snapper_zypp_plugin.cc
```diff
@@ -302,7 +302,6 @@ set<string> SnapperZyppPlugin::get_solvables(const Message& msg, Phase phase) {
 
     if (json_object_get_type(steps) == json_type_array) {
 	size_t i, len = json_object_array_length(steps);
-	printf("steps: %zu\n", len);
```
jreidinger: maybe document in top of plugin that no such calls are allowes otherwise this can happen ( together with bsc) ?

mlandres:
```
    sendback = sys.stdout
    sys.stdout = sys.stderr
    print "debug...."
    print >>sendback, "plugin responses..."
```
My py-plugins immediately redirect stdout to stderr after having made a copy of stdout which is explicitely used to send the plugin responses.

mvidner: Thanks!
Instead I made a forwarding-zypp-plugin helper that catches the bug at `make check` time.

zypp-plugin/forwarding_zypp_plugin.cc
```diff
@@ -0,0 +1,71 @@
...
+
+#include <string>
+using namespace std;
+
+#include <boost/process.hpp>
+namespace bp = boost::process;
+
+#include "zypp_plugin.h"
+
+class ForwardingZyppPlugin : public ZyppPlugin {
```
aschnell: Please explain the purpose of the class.

mvidner: Ah, you're right. See above now.

PR #523, opened 2020-02-24
Add flag to disable install of systemd files

TBK:


TBK:
CI fails due to docker timeout in the ubuntu job - https://travis-ci.org/openSUSE/snapper/jobs/654345031

PR #527, opened 2020-03-06
feat: add initial (and experimental) support for ZFS

johanfleury:
On config creation, a new dataset is created (`.snapshots`) that will
store snapshots information and a symlink to the snapshot in the `.zfs`
directory.  This dataset is destroyed on config deletion.

Due to a ZFS “limitations”, snapshots can only be created read-only and
cannot be derived from a parent.

This is based on the initial idea and work of Daniel Sullivan
(mumblepins).

Implements #145.


aisbergg:
I really like to see ZFS support implemented in snapper. There are a few snapshot programs, that can be used with ZFS, but non of which are as convenient:

- https://github.com/yboetz/pyznap
- https://github.com/grantma/zsnapd
- https://github.com/zfsnap/zfsnap.org

I don't code in C++, but I can offer my help testing the feature.

i2:
Is there any update on this PR?

Username404-59:
ZFS support would be great

NicholasFlamy:
> There’s no interest from upstream. If you need automated zfs snapshot, either fork the project or use something else.

I figured. It's sad but I understand. Thank you for trying.

PR #539, opened 2020-06-01
provide a doc configure option

dlan17:
Give user an option to enable/disable building man helps

Signed-off-by: Yixun Lan <dlan@gentoo.org>


dlan17:
I'm not sure if you guys like this idea or not ...

The motivation is that users may not want to build man/doc, providing this option give them a choice to disable it.

PR #540, opened 2020-06-18
- special rollback for transactional server (bsc#1172273)

aschnell:
This PR specializes the rollback command for transactional servers. The detection of transactional servers is done automatic but can be overruled by a global command line option.

The error handling of command line options is also change to use exception reducing the LOC significantly.

See https://trello.com/c/7wOPtMOn/1892-3-ostumbleweed-p5-1172273-snapper-rollback-considered-harmful-in-a-transactional-installation and https://bugzilla.suse.com/show_bug.cgi?id=1172273.

PR #541, opened 2020-06-18
Workaround for a Travis Tumbleweed build error

lslezak:
- Related to https://travis-ci.org/github/openSUSE/snapper/jobs/687217532
- Explicitly install "util-linux", the automatically selected  "busybox-util-linux" creates conflicts later
- It looks like "ruby" has some weaker dependency and installing "busybox-util-linux" is enough, later some other tool needs full "util-linux" and it conflicts with the already installed "busybox-util-linux"


aschnell:
Included in PR #540.

PR #543, opened 2020-06-19
Improve openSUSE Travis builds

lslezak:
- Do not mix the Leap and Tumbleweed YaST repositories
- Pin the openSUSE version to 15.1 (the "latest" might be switched to 15.2 anytime in the future causing inconsistenty with the used 15.1 YaST repository)
- Later we can run 15.1 and 15.2 in parallel if needed or just switch from 15.1 to 15.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,15 @@ services:
 
 # define the build matrix
 env:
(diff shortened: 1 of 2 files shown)


aschnell:
LGTM

PR #544, opened 2020-07-13
- added error handing for failed ambit detection (bsc#1174038)

aschnell:
For https://bugzilla.suse.com/show_bug.cgi?id=1174038.

The current btrfs default snapshot can be unknown to snapper. In that case the ambit detection fails. Print an error message in that case.


client/snapper.cc
```diff
@@ -1067,6 +1067,14 @@ command_rollback(cli::GlobalOptions& global_options, ProxySnappers* snappers, Pr
 
     if (global_options.ambit() == cli::GlobalOptions::Ambit::AUTO)
     {
+	if (previous_default == snapshots.end())
+	{
+	    cerr << _("Cannot detect ambit since default subvolume is unknown.") << '\n'
+		 << _("This can happen if the system was not setup for rollback.") << '\n'
```
shundhammer, line 1073: "set up" (two words)

shundhammer:
Just the one minor English issue, otherwise LGTM.

client/snapper.cc
aschnell: Fixed.

PR #551, opened 2020-08-31
Improve snapperd dbus activation

lnussel:
Dbus needs to know the systemd unit to activate in order to avoid using
a setuid launch helper. Ie this makes snapperd work in paranoid
permissions mode.


aschnell:
The new file must also be added to snapper.spec.in file.

lnussel:
updated

aschnell:
Thanks.

PR #579, opened 2020-10-10
Undefine stdout and stderr in SystemCmd.h.

ericonr:
musl's stdio.h provides the stdout and stderr macros, which conflict
with the method definitions in SystemCmd and cause compilation errors.

Taken from #220


aschnell:
In PR https://github.com/openSUSE/snapper/pull/594 I renamed the functions.

PR #580, opened 2020-10-10
Avoid glibc-specific time extensions.

ericonr:
* Use mktime instead of timelocal
* Replace glibc specific time format string
* Inline private __isleap() macro from glibc

Taken from #220


aschnell:
Thanks, looks good.

PR #581, opened 2020-10-10
Always use XSI-compliant strerror_r.

ericonr:
This makes the program simpler and employs feature test macros properly,
since they should be defined before the inclusion of any headers.

### DIFF
--- a/snapper/AppUtil.cc
+++ b/snapper/AppUtil.cc
@@ -20,6 +20,8 @@
  * find current contact information at www.novell.com.
  */
 
+#define _XOPEN_SOURCE 700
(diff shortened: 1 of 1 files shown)


ericonr:
Doing this might be a bit more complicated than I expected, unless I split it into an additional file. Would that be okay? The issue here is that on musl this won't compile, because those macros aren't defined (they should be defined by the application, instead).

aschnell:
OK, it seems as if I never read the documentation about feature_test_macros. And please use something like ```int r = strerror_r(...)``` so that the compiler gives an error if the return type is ```char*```.

aschnell:
See https://github.com/openSUSE/snapper/pull/605.

ericonr:
Heh, great solution. Thank you for looking into these!

PR #582, opened 2020-10-10
Use default value if _SC_GETPW_R_SIZE_MAX is unavailable.

ericonr:
sysconf may return -1 if the value is unlimited.

Based on #220


aschnell:
Thanks, look fine.

PR #583, opened 2020-10-10
Search for GNU compatible gettext implementation

ericonr:
snapper uses a GNU specific symbol in gettext (_nl_msg_cat_cntr), which
is not provided by musl's gettext implementation.

Based on #220

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,9 @@ fi
 
 AC_CHECK_HEADER(acl/libacl.h,[],[AC_MSG_ERROR([Cannout find libacl headers. Please install libacl-devel])])
 
+AM_GNU_GETTEXT_VERSION([0.18.1])
+AM_GNU_GETTEXT([external])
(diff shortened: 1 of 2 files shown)


aschnell:
I would prefer to drop the own Regex class and use C++ regexes instead. I plan to look into that this week.

aschnell:
Now C++11 regexes are used, see https://github.com/openSUSE/snapper/pull/595.

Unfortunately since the zypp-plugin used the own Regex class and potentially problems during upgrades have to be avoided, the Regex class is still present. I assume that zypp-plugin is only used on SUSE so you should be fine by simply remove Regex.{cc,h} from Makefile.am.

ericonr:
@aschnell would it be okay to make a PR to make the plugin optional? Either way, thank you!

aschnell:
There is already the ```--disable-zypp``` switch.

Issue #659, opened 2021-06-15
securelibdir determined incorrectly on riscv


GreatEmerald:
We have a report that on the riscv architecture, the install script misdetects the library path for pam_snapper.so and puts it into a directory with the ABI name rater than in a lib directory. I'm guessing the `$(libdir)` is wrongly determined here:

https://github.com/openSUSE/snapper/blob/23a6bdfefa49b737557b04237d29e93fc3b4f435/pam/Makefile.am#L11

It gets installed into `/lp64d/security/`, which does not exist.

Downstream bug report: https://bugs.gentoo.org/796179

aschnell:
So is libdir = /usr/lp64d and the correct dir for pam plugins is /lib64?

GreatEmerald:
From what I can tell, libdir = /lp64d, no /usr. I'm not entirely certain what the correct dir is, it's either lib or lib64.

aschnell:
Also see https://github.com/linux-pam/linux-pam/issues/256.

aschnell:
The reference to linux-pam shows that an automatic detection of the correct path is no far provided.

I have added the configure option ```--with-pam-security``` that can be used if the default does not fit.

PR #660, opened 2021-06-16
- added configure option for location of PAM module

aschnell:
See https://github.com/openSUSE/snapper/issues/659.


DimStar77:
pam-devel brings a macros.pam, defining the corresponding value as `%{_pamdir}` - valid since SLE15 (with update channe)

PR #677, opened 2021-11-01
Add musl libc support

leonardohn:
The musl libc provides libintl (similar to glibc) but does not use the same internals, so snapper fails to link due to missing _nl_msg_cat_cntr symbol. As the Regex class is signed as obsolete, one simple solution for this issue is to remove this class.


aschnell:
That breaks the ABI and thus the major version of libsnapper has to be increased. I do not want to do that just for a tiny cleanup. You can simply remove the line in Makefile.am during build. Also see https://github.com/openSUSE/snapper/pull/595.

PR #680, opened 2021-11-24
Add Bash completion

ianjazz246:
Bash completion based on #278


aschnell:
Looks fine so far.

I merge it now and do the additions to spec and dsc files.

PR #689, opened 2022-01-28
add glibc-locale buildrequires for suse like distros

dirkmueller:
The testsuite uses locales from non glibc-locale-base so we need the explicit buildrequires.

Also cleanup the spec file a little while being at it.


dirkmueller:
whoops, this is already in.

PR #706, opened 2022-04-18
Manpage: cron -> systemd

tacerus:
Signed-off-by: Georg Pfuetzenreuter <mail@georg-pfuetzenreuter.net>

Hi,

I think this line in the manpage could use an update, as I could not find cron-jobs for snapper on a new Tumbleweed install - only the systemd timers seem to exist.

Best,
Georg

### DIFF
--- a/doc/snapper.xml.in
+++ b/doc/snapper.xml.in
@@ -98,7 +98,7 @@
       <para>Next to manual snapshot creation, snapshots are also created automatically.</para>
       <itemizedlist>
 	<listitem>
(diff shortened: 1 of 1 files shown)


aschnell:
Some distributions may still use cron jobs. So please mention both.

aschnell:
Thanks.

tacerus:
Thank you, shall I update the `gh-pages` branch as well, or is it generated somehow?

aschnell:
I generate the gh-pages once in a while. Can do so now.

PR #732, opened 2022-06-20
Moved logrotate files from user specif directory /etc/logrotate.d to …

schubi2:
…vendor specif directory /usr/etc/logrotate.d.

### DIFF
--- a/package/snapper.changes
+++ b/package/snapper.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Aug 30 10:35:28 UTC 2022 - Stefan Schubert <schubi@suse.com>
+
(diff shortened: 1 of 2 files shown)


aschnell:
The .rpmsave handling is missing. And should files in /usr/etc really be marked as "noreplace"?

Please add a changelog entry.

schubi2:
> The .rpmsave handling is missing. And should files in /usr/etc really be marked as "noreplace"?
Files in /usr/etc are vendor files which can be changed by an RPM update. User changed files are still in /etc directory.
I have fixed the rest. Thank you !

aschnell:
Looks good. Thanks.

PR #743, opened 2022-09-02
Fix build with GCC 12 (and musl)

thesamesam:
```
DBusMessage.h:239:34: error: 'time_t' has not been declared
  239 |     Hihi& operator>>(Hihi& hihi, time_t& data);
```

Include <ctime> to grab time_t.

Bug: https://bugs.gentoo.org/862094
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/dbus/DBusMessage.h
+++ b/dbus/DBusMessage.h
@@ -27,6 +27,7 @@
 
 #include <dbus/dbus.h>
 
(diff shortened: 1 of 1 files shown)


aschnell:
Looks fine. Thanks.

PR #760, opened 2022-12-21
Call delete-snapshot plugin earlier

lnussel:
Call the delete-snapshot plugin before actually removing any files, in case it still needs to fetch information from the snapshot.

### DIFF
--- a/snapper/Snapshot.cc
+++ b/snapper/Snapshot.cc
@@ -764,6 +764,8 @@ namespace snapper
 	    snapshot->isActive())
 	    SN_THROW(IllegalSnapshotException());
 
+	Hooks::delete_snapshot(snapper->subvolumeDir(), snapper->getFilesystem(), *snapshot);
+
 	snapshot->deleteFilesystemSnapshot();
 
 	SDir info_dir = snapshot->openInfoDir();
@@ -791,8 +793,6 @@ namespace snapper
 
 	SDir infos_dir = snapper->openInfosDir();
 	infos_dir.unlink(decString(snapshot->getNum()), AT_REMOVEDIR);
-
-	Hooks::delete_snapshot(snapper->subvolumeDir(), snapper->getFilesystem(), *snapshot);
 
 	entries.erase(snapshot);
     }


aschnell:
Did you test this with the grub plugin?

lnussel:
*gmbl* nope. that one iterates over all configs and would then of  course pick up the to-be-deleted snapshot. So I guess we either we implement yet another hook, call the grub hack explicitly later or we touch a file that marks the snapshot as to be deleted. What's your preferred pain? :)

aschnell:
The grub script shows that some scripts prefer to be called after the deletion.

Could we pass the scripts (the new once starting with a number) the option "--pre-action" / "--post-action" and simply call them twice? Or would that break to much?

aschnell:
Something like https://github.com/openSUSE/snapper/commit/292a50d4206a974f68be5fc5433cc85fac1404b6 should work.

The delete-snapshot plugin is called before the action. Also --stage=pre-action/post-action is passed to the plugins (
additional options for plugins have been allowed). When needed the plugins can then also be called a second time.

Untested.

lnussel:
That adds the complication that the plugins now have to parse options in addition. How about just calling the command eg. 'delete-snapshot-pre', 'create-snapshot-post' etc?

lnussel:
Current plugin I use https://build.opensuse.org/package/view_file/devel:microos:systemd-boot/kernel-install-openSUSE/10-kernel-install.snapper?expand=1

aschnell:
Yes, that is indeed better since it should not break any existing scripts.

lnussel:
I doubt that there are any other than mine but making it simpler would certainly help me :)

aschnell:
Next try: https://github.com/aschnell/snapper/commit/9980f8a54dfa9a0f595a91e4e09bb40174e77334

lnussel:
LGTM. IMO you can drop the compat calls though :-) No need to call scripts twice. Alternatively we could also leave the old name in the old place and just add a single pre/post. Anyway fine with me

aschnell:
Thanks. Replaced by https://github.com/openSUSE/snapper/pull/761.

PR #765, opened 2023-01-04
Replace transitional %usrmerged macro with regular version check (boo#1206798)

lnussel:
Replace transitional %usrmerged macro with regular version check (boo#1206798)


aschnell:
Thanks.

PR #767, opened 2023-01-04
Add build detection of libbtrfsutil and use for 'set default subvolume'

kdave:
A preview for libbtrfsutil support. Configure-time detection, sample code that's in an ifdef as it does not share much except of function arguments, which would be case for most implemented operations. Before implementing the rest I'd need feedback if this the way to do it or feel free to use it, the util implementations can be added incrementally.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -179,6 +179,14 @@ PKG_CHECK_MODULES(XML2, libxml-2.0)
 PKG_CHECK_MODULES(JSONC, json-c, [], [AC_MSG_WARN([Cannot find json-c. Please install libjson-c-devel])])
 PKG_CHECK_MODULES(ZLIB, zlib)
 
+AC_CHECK_LIB(btrfsutil, btrfs_util_strerror)
+AC_CHECK_HEADERS([btrfsutil.h])
+
+# Conditional support for libbtrfsutil
+if test "x$ac_cv_lib_btrfsutil_btrfs_util_strerror" = "xyes"; then
+	PKG_CHECK_MODULES(LIBBTRFSUTIL, libbtrfsutil)
+fi
+
 AC_CHECK_HEADER(acl/libacl.h,[],[AC_MSG_ERROR([Cannout find libacl headers. Please install libacl-devel])])
 
 AC_SUBST(VERSION)
--- a/snapper/BtrfsUtils.cc
+++ b/snapper/BtrfsUtils.cc
@@ -44,6 +44,9 @@
 #include "snapper/AppUtil.h"
 #include "snapper/BtrfsUtils.h"
 
+#ifdef HAVE_LIBBTRFSUTIL
+#include <btrfsutil.h>
+#endif
 
 namespace snapper
 {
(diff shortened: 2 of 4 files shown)


snapper/Makefile.am
```diff
@@ -61,9 +61,9 @@ libsnapper_la_SOURCES +=				\
 	Selinux.cc		Selinux.h
 endif
 
-libsnapper_la_CPPFLAGS = $(XML2_CFLAGS) $(ZLIB_CFLAGS)
+libsnapper_la_CPPFLAGS = $(XML2_CFLAGS) $(ZLIB_CFLAGS) $(LIBBTRFSUTIL_CFLAGS)
 libsnapper_la_LDFLAGS = -version-info @LIBVERSION_INFO@
-libsnapper_la_LIBADD = -lboost_thread -lboost_system $(XML2_LIBS) -lacl $(ZLIB_LIBS)
+libsnapper_la_LIBADD = -lboost_thread -lboost_system $(XML2_LIBS) -lacl $(ZLIB_LIBS) $(LIBBTRFSUTIL_LIBS)
```
aschnell, line 66: Those additions should only be done if ENABLE_BTRFS is set.

aschnell:
In general OK.

Should btrfsutil be used on Tumbleweed then? In that case some conditional BuildRequires is also needed in snapper.spec.in. I looked into that yesterday and can add that.

configure.ac
```diff
@@ -179,6 +182,14 @@ PKG_CHECK_MODULES(XML2, libxml-2.0)
 PKG_CHECK_MODULES(JSONC, json-c, [], [AC_MSG_WARN([Cannot find json-c. Please install libjson-c-devel])])
 PKG_CHECK_MODULES(ZLIB, zlib)
 
+AC_CHECK_LIB(btrfsutil, btrfs_util_strerror)
+AC_CHECK_HEADERS([btrfsutil.h])
```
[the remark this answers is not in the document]
kdave: Yes, though my idea was to keep the AC_CHECK_LIB right before the PKG_CHECK_MODULES, not the first one (line 110).

snapper/Makefile.am
kdave: Ok, moved under "if ENABLE_BTRFS".

kdave:
Branch updated. Yes, build test on tumbleweed would be good, I don't know which macro/version to use so please add it.

kdave:
I've implemented the remaining methods where the libbtrfsutil provides the functionality, should I merge it into one commit or do you want to split the build support and implementation?

aschnell:
Just add using more methods of libbtrfsutil here. Please also add a changelog entry. Thanks.

kdave:
Updated and pushed.

snapper/BtrfsUtils.cc
```diff
@@ -67,30 +70,79 @@ namespace snapper
...
+		inherit->num_excl_copies = 0;
+		inherit->qgroups[0] = qgroup;
+		util_inherit = (struct btrfs_util_qgroup_inherit *)inherit;
+	    }
+#endif
+	    err = btrfs_util_create_snapshot_fd2(fd, fddst, name.c_str(), flags, NULL, util_inherit);
+	    if (err && errno != ENOTTY && errno != EINVAL)
+		throw runtime_error_with_errno("btrfs_util_create_snapshot_fd2() failed", errno);
+
+	    /* No BTRFS_IOC_SNAP_CREATE fallback */
```
aschnell, line 144: AFAIS the fallback does not use ```BTRFS_IOC_SNAP_CREATE``` since the second ```#endif``` below is too late.

Whether the fallback is needed is another question (libbtrfsutil seems to be a few years younger than ```BTRFS_IOC_SNAP_CREATE_V2```).

PR #771, opened 2023-01-19
Implement BtrfsUtils::get_subvolume using libbtrfsutil

kdave:
The helper from btrfs_subvolid_resolve from libbtrfs can be reimplemented by means of libbtrfsutil. btrfs_util_subvolume_path_fd is practically equivalent to btrfs_subvolid_resolve i.e. requires the privileged ioctl. There's another way available to all users but is slow as it needs to iterate overall subvolumes. This is implemented as fallback.


aschnell:
Thanks.

PR #780, opened 2023-02-07
Add zsh completion

### LINKED ISSUE #623: RFE: Add zsh completions
Snapper has plenty of command line "switches" and to make them more discoverable and enter them easily, I would suggest adding a completion file for zsh. But why zsh if bash is the default shell for most distributions?
Well there is #278 ,  so why not add one for zsh users.

Freed-Wu:
Fix #623, similar with #680


aschnell:
I have added the --no-headers option so please use it here. Apart from that please add a changelog entry
and update the file lists in snapper.spec.in file and dists/debian/snapper.install.

aschnell:
Thanks.

I assume for zsh the site-functions directory itself must be package otherwise the build service will complain that the directory does not belong to any package. We'll see.

PR #832, opened 2023-08-24
Implement BtrfsUtils::get_id using libbtrfsutil

kdave:
The get_id() member can be also implemented using libbtrfsutil. A separate conditional implementation is added and the #ifdefs updated that into account.


aschnell:
Thanks, looks fine.

PR #837, opened 2023-09-16
Update useradd script to prevent warnings

daniel-fahey:
Not sure if this script is still recommended to use as it's not been modified in nearly a decade.

1. Modified the `useradd` command by removing the `-m` option to prevent unnecessary warnings about the home directory already existing. The creation of the home directory is managed separately in this script using Btrfs subvolumes.
   
2. Added a command to copy the standard skeleton files from `/etc/skel` to the new user's home directory, ensuring that new users receive the necessary setup files despite the custom home directory creation process with the `create-config` command.

3. Corrected the `chown` command syntax to use the more modern `:` separator between the user and group names, getting rid of another warning.

I kept the (old school) style as is. Thanks for snapper!


aschnell:
The script is likely bare used (esp. with the DRYRUN=1 setting).

Anyway, I still get a warning about the home directory already existing unless I add -M.

PR #838, opened 2023-09-19
PR #837 follow-on (scripts/pam_snapper_useradd.sh) `useradd -M`

daniel-fahey:
Running `useradd` without options may still try to create the user's home directory depending on system configuration, `useradd -M` attempts to override this


daniel-fahey:
Good Morning Arvin

No worries, thanks so much for trying it out yourself too.

PR #861, opened 2023-12-22
Add codespell support (config, workflow to alert on new typos) + make it fix typos

yarikoptic:


aschnell:
Thanks, could you add a link to the documentation in codespell.yml?

.codespellrc
```diff
@@ -0,0 +1,5 @@
+[codespell]
```
yarikoptic, line 1:
```suggestion
# Ref: https://github.com/codespell-project/codespell#using-a-config-file
[codespell]
```

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,22 @@
+---
```
yarikoptic, line 1:
```suggestion
# Configure via .codespellrc in the root of the project
---
```

yarikoptic:
Added references in the config and in the workflow

PR #1015, opened 2025-05-29
Fix timer description to match actual hourly cleanup frequency

RiverOnVenus:
The snapper-cleanup.timer runs the cleanup every hour (set by OnUnitActiveSec=1h). But its description says "Daily Cleanup of Snapper Snapshots," 

This can confuse people who use snapper for the first time and see the timer status. They might think cleanup only happens once a day.


aschnell:
Thanks.

Issue #1021, opened 2025-06-17
snbk can't find snapshots on mounted subvolume


kermeat:
snbk probe can't find snapshots if target directory is on a mounted subvolume.

backup config:
```
{
    "config": "home",
    "target-mode": "local",
    "automatic": true,
    "source-path": "/home",
    "target-path": "/.backup/home"
}
```
mount point:
```
TARGET   SOURCE                    FSTYPE OPTIONS
/.backup /dev/sdc[/backup/snapper] btrfs  rw,relatime,compress=zstd:13,space_cache=v2,subvolid=289,subvol=/backup/snapper
```
After  the transfer is completed `snbk list` shows all snapshots as missing.
If I change target-path in the config to the top level subvolume (`/mnt/backup/snapper/home`) the snapshots become valid.

version 0.12.2

PR #1022, opened 2025-06-21
fix snbk probe_target issue with non top-level subvolumes

kermeat:
fix for issue https://github.com/openSUSE/snapper/issues/1021


aschnell:
Thanks, looks fine.

Issue #194, opened 2015-09-07
How to restore snapper metadata for snapshots after disk failure/replacement (using btrfs send/receive)


kostgr:
I'd like to implement backup of my system (root and home) using btrfs send/receive to an external drive (also btrfs of cause).

If I have to restore the data from the backup after disk failure/replacement, I restore the disk layout and do something like this:
btrfs /mnt/backup/path/to/last/snapshot send | btrfs receive /

But after that, I have no snapshots and cannot go further back, if I have to.

The better case could be, to restore all the snapshots from snapper with the metadata (using some script) and then use snapper as if the disk were never replaced.

So how could I backup the snapshots metadata i.e. restore them?

aschnell:
With the btrfs send/receive command you only restore the last snapshot to your root filesystem. In that case you must recreate the snapper config for root. Since you have also restored /etc/sysconfig/snapper you first have to remove root there and then run "snapper create-config /". (Note: I haven't tested that, so maybe some steps are missing.)

For restoring all snapshots there is a project https://github.com/doudou/snapsync. Unfortunately I didn't have time to try that yet so I cannot comment on its features and stability.

kostgr:
I've configured the snapsync a couple of months before. It has done a good job of sending incremental backups via btrfs send/receive with management for snapshots to be kept on backup drive like the snapper does for local snapshots, but hadn't got a chance to test restoration till now.

I've had to restore snapper configuration today (reinstalled OS on another partition). I've found out, that I cannot simply use the command `snapper -c DataStore_hdd create-config /media/DataStore`, because the partition /media/DataStore already has .snapshot-directory. So I had to restore the configuration manually.
That is what I've done to restore snapper configuration:
1. I had to modify /etc/sysconfig/snapper like this:
`SNAPPER_CONFIGS="root"` to `SNAPPER_CONFIGS="root DataStore_hdd"`
2. I had to copy the configuration file from the last snapshot like this:
`cp /media/usbDataStore/backups/root/541/snapshot/etc/snapper/configs/DataStore_hdd /etc/snapper/configs/`

It is actually better, then configuring it from scratch, but it was a pity that I couldn't find the exact commands on the web.

Now it is time to restore snapper configuration... I'll write about my experiences as soon as I've ended the quest.

PR #1072, opened 2025-12-10
Add snapshots restoring support to snbk

jameslai-dev:
The pull request adds a snapshot restoring support to `snbk`.
This allows users to transfer snapshots from the backup target back to the source using the following new command:

```bash
snbk restore [numbers]
```

This PR provides a foundational piece toward resolving issue #194.
I also personally need this feature for restoring server systems in the event of a disaster.

There is one limitation I have not been able to resolve: after restoring snapshots, `snapper` is not showing the restored snapshots unless the `snapperd` service is restarted.
I would appreciate any advice on how to address this. If there is anything that needs to be changed in this PR, please let me know!

#### Btrfs Send Parent Finding

The algorithm for finding the Btrfs send parent during restoration is slightly different from that used during snapshot transfer.

When transferring snapshots to the backup target, `snbk` requires an explicit parent snapshot to perform an incremental backup. However, these parent snapshots may have been removed by timeline cleanup.
To reduce disk usage, the nearest previous available snapshot is used as the send parent if the explicit parent does not exist.

I would like to open another PR to enhance and unify the send parent finding algorithm for both transferring and restoring with a tree-based searching. I would appreciate your feedback.

#### Shell Module

A `shell1` parameter is added to the `shellify_pipe` function. I think this change is required for restoring snapshots from an SSH backup target.

#### Potential Duplicated Implementations

The implementations for transferring and restoring snapshots share duplicated logic.
If this PR is accepted, I would like to submit another PR to unify these implementations to reduce maintenance overhead.

#### Coding Style

From viewing the codebase, it appears that a line width of approximately 90 characters is used.
I have tried to follow the existing style in this PR. If any adjustments are needed or if a specific formatter is used,
please let me know, and I will reformat accordingly.

### DIFF
--- a/package/snapper.changes
+++ b/package/snapper.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Dec 17 12:57:19 UTC 2025 - Cheng-Ling Lai <jamesljlster@gmail.com>
+
+- Add snapshots restoring support to snbk
+  (gh#openSUSE/snapper#1072)
+
+-------------------------------------------------------------------
 Tue Nov 18 10:23:16 CET 2025 - aschnell@suse.com
 
 - add CAP_MKNOD in snapper-backup.service


jamesljlster:
> Thanks a lot for the patch. It look very good.

No problem! It’s my pleasure to contribute.

> One problem is that snapperd will not know about the restored snapshots (if it is currently running). That should be documented.

Got it. I will add some documentation for this in a future commit.

> May you mind explaining your use case?

Sure! We use Ubuntu 24.04 on Btrfs for our production server systems. I have successfully restored misconfigured settings several times with Btrfs snapshots, which are really helpful.

Recently, we started doing full system backups for our production systems. Since Btrfs is used, I prefer making backups using the method natively supported by Btrfs  —  Btrfs send/receive. So that I don’t need to worry about any filesystem-level inconsistency between the source and the backups.

We store the system backups to iSCSI disks provided by our storage server. With this setup:

1. If the hardware RAID on a production server fails, we can replace the storage and restore the system from the backups stored on the storage server.
2. If the production server hardware is faulty, we can restore the system to a temporary VM on another host, ensuring important services continue without a long downtime.
3. If the storage server fails, we can create backups again from the source snapshots.

After restoring the system, I need to make sure all snapshots are fully restored. Otherwise, the backups might be removed when the scheduled snapper-backup service is triggered. This is the reason why I created this pull request.

For 2 and 3 is the reason why I would like to propose a new Btrfs send parent finding algorithm. The parents of snapshots may be cleaned up by the timeline service. Snapshots without explicit parents won’t be sent incrementally, leading to additional disk space usage overhead.

Please let me know if you want to know more about the details.

aschnell:
When running restore and there are several backup configs for one snapper config the code tries to restore the snapshot several times. I think a check for ```backup_configs.size() == 1``` is needed in ``` command_restore()```. So that the user has to provided the --backup-config option (if there are several).

aschnell:
Thanks for describing your use case. Are you restoring snapshots of the root file system or some data file system? If for the root file system how do you manage the various different UUIDs (e.g. of the root btrfs file system in fstab)?

jamesljlster:
Sure! I've updated the changelog. Please take a look.

PR #1101, opened 2026-02-05
Prevent manually transferring a read‑write snapshot

jameslai-dev:
This pull request adds a check in `TheBigThing::restore`.
The check prevents `snbk` from creating an invalid snapshot on the target device when a user manually transfers a read-write snapshot.


aschnell:
Thanks.

PR #1126, opened 2026-04-19
[ci] Update GitHub Actions to latest versions

gruenich:
Fix a couple of deprecation warnings that Node20.js will stop working in June


aschnell:
Thanks. Just a question: Where did you see the deprecation warnings? I could not find them.

gruenich:
One of the Codespell Actions showed the deprecation warning, see https://github.com/openSUSE/snapper/actions/runs/23656634847.

I found the others when Git grepping for similar occurrences. You are right, it is weird that the other CIs did not trigger deprecation warnings.

PR #1132, opened 2026-05-14
snapper-sync: skip placeholder when source-config has no snapshots

### LINKED ISSUE #1131: snapper-sync creates a literal '.snapshots/null' directory when a source-config has no snapshots yet
## Summary

`scripts/snapper-sync` (introduced in #1128, shipped as part of `snapper-backup` since `v0.13.1`) creates a placeholder directory named literally `null` under `<subvolume>/.snapshots/` whenever a configured backup-config maps to a source-config that has no snapshots in the output of `snbk list`.

This affects, for example, a freshly configured `snbk` backup-config whose first transfer has not yet completed, or any source-config that legitimately has zero snapshots at the moment `snapper-sync.service` runs.

## Reproducer

The bug is entirely deterministic in the `jq` pipeline used inside `scripts/snapper-sync`. Running the same expression as the script with empty mock data shows the issue without needing a real snapper setup:

```sh
echo '{"snapshots":[]}' \
  | jq -r '[.snapshots[] | select(.config == "root")] | max_by(.number) | .number'
# prints: null
```

The script then takes that `null` literally and runs `mkdir <subvol>/.snapshots/null`. With my own mock data wrapped around the script's body I reproduced it directly:

```text
=== Scenario: empty list ===
max_num='null' target_path='./test-root/.snapshots/null'
./test-root/.snapshots/null created.

Result: null
```

A real-world reproducer is roughly:

1. `snbk create-config backup_root` (a fresh backup-config that has never run).
2. `systemctl start snapper-sync.service`.
3. `ls /.snapshots/` — observe an unexpected `null` entry.

## Root cause

`scripts/snapper-sync` lines 45-48:

```sh
max_num=$(
  echo "$snapshots" |
  jq -r "[.[] | select(.config == \"$source_config\")] | max_by(.number) | .number"
)
```

When the filtered array is empty, `max_by` returns `null`, `.number` on `null` is `null`, and `jq -r` prints the string `null`. The script then unconditionally falls through to:

```sh
target_path="$subvol_path/.snapshots/$max_num"
if [ -d "$target_path" ]; then
  echo "$target_path exists."
else
  mkdir "$target_path"
  echo "$target_path created."
fi
```

…so the literal directory `<subvolume>/.snapshots/null` is created. `snapper`'s own `nextNumber()` only counts directories whose names match `[1-9][0-9]*`, so the placeholder is silently ignored by snapper but remains on disk as unexpected cruft that the user has to clean up manually.

## Additional small finding while debugging

`shellcheck` flags `scripts/snapper-sync:28`:

```sh
echo $outputs | jq ...
```

The unquoted expansion of `$outputs` collapses whitespace; this happens to be safe for JSON only by coincidence and is otherwise a `SC2086` warning. Worth fixing in the same change.

## Proposed fix

- Make the `jq` expression emit no output (rather than `null`) when there are no matching snapshots, and additionally `select(.number != null)` to defend against entries that legitimately have no number.
- Skip the iteration with a clear log message when `max_num` ends up empty.
- Quote `"$outputs"` to silence `SC2086`.

I have a fix prepared and will open a PR shortly.

officialasishkumar:
## Summary

- Stop `scripts/snapper-sync` from creating a literal `.snapshots/null` directory when a source-config has no snapshots yet (e.g. a freshly configured `snbk` backup-config whose first transfer has not yet run, or a source-config that legitimately has zero snapshots when `snapper-sync.service` fires).
- Quote `"$outputs"` in the JSON pass-through to jq, clearing the only `shellcheck` `SC2086` warning the script otherwise emits.

Fixes #1131.

## Root cause

For each backup-config, the script picks the highest snapshot number from the `snbk list` output and `mkdir`s `<subvolume>/.snapshots/<n>` as a placeholder so that snapper's next snapshot number is forced past `<n>`. When the filtered array is empty, `max_by` returns `null`, `.number` on `null` is `null`, and `jq -r` prints the literal string `null`. The script then runs `mkdir <subvolume>/.snapshots/null`. Snapper's own `nextNumber()` only counts entries matching `[1-9][0-9]*`, so the placeholder is silently ignored on the snapper side but stays on disk as cruft the user has to clean up manually.

```sh
echo '{"snapshots":[]}' \
  | jq -r '[.snapshots[] | select(.config == "root")] | max_by(.number) | .number'
# -> null
```

## Fix

- Wrap the jq expression in `if length == 0 then empty else (max_by(.number) | .number) end` after first filtering out entries whose `.number` is itself `null`. When there are no matching snapshots, jq emits nothing, so `$max_num` ends up an empty string instead of the literal `null`.
- Skip the iteration with a clear `'No snapshots found for source-config '<name>'; skipping.'` message when `$max_num` is empty.
- Quote `"$outputs"` (SC2086).

## Test plan

I wrote a small harness around the patched script body and ran six scenarios:

```text
=== Scenario: empty list ===
No snapshots found for source-config 'root'; skipping.
Files in root1/.snapshots: ''

=== Scenario: single snapshot ===
./root2/.snapshots/5 created.
Files in root2/.snapshots: '5'

=== Scenario: existing placeholder ===
./root3/.snapshots/42 exists.
Files in root3/.snapshots: '42'

=== Scenario: two backup-configs share source ===
./root4/.snapshots/9 created.
./root4/.snapshots/9 exists.
Files in root4/.snapshots: '9'

=== Scenario: no matching snapshots ===
No snapshots found for source-config 'root'; skipping.
Files in root5/.snapshots: ''

=== Scenario: snapshot entry with null number ===
./root6/.snapshots/12 created.
Files in root6/.snapshots: '12'
```

All scenarios behave correctly: empty/no-match cases skip cleanly without creating a literal `null` directory, normal cases still create the right placeholder, and shared-source backup-configs still report `exists` on the second pass.

`shellcheck -s sh scripts/snapper-sync` is clean on the patched script (it reports `SC2086` on the unpatched version).

- [x] Reproduced the bug end-to-end with the script's own jq pipeline.
- [x] Patched script passes the six-scenario harness above.
- [x] `shellcheck -s sh scripts/snapper-sync` reports no findings.
- [x] No behavior change for source-configs that have snapshots.

### DIFF
--- a/scripts/snapper-sync
+++ b/scripts/snapper-sync
@@ -25,7 +25,7 @@ config_mapping=$(
 
 if outputs=$(snbk --machine-readable json list); then
   snapshots=$(
-    echo $outputs |
+    echo "$outputs" |
     jq --argjson config_mapping "$config_mapping" '
       .snapshots | . |= map(. + {config: $config_mapping[.name]})
     '
@@ -46,6 +46,14 @@ echo "$config_mapping" | jq -r 'keys[]' | while read -r backup_config; do
     echo "$snapshots" |
     jq -r "[.[] | select(.config == \"$source_config\")] | max_by(.number) | .number"
   )
+
+  # Nothing to synchronize for this source-config (e.g. no snapshots have been transferred
+  # and the source has none either). Skip to avoid creating a literal '.snapshots/null'
+  # placeholder directory.
+  if [ "$max_num" = "null" ]; then
+    echo "No snapshots found for source-config '$source_config'; skipping."
+    continue
+  fi
 
   # Query the source subvolume path
   subvol_path=$(


jameslai-dev:
I would like to leave a few comments here.
In my opinion, there is no logical difference between:

1. Checking if `max_num` is `null`.
2. Setting `max_num` to an empty string when no snapshot is present, then checking if it is empty.

However, the first option keeps the `jq` query simpler.

officialasishkumar:
Simplified the jq expression in `91cb820` so it relies on `// empty` and keeps the existing empty-value guard. I also reran `sh -n scripts/snapper-sync` and a small fake `snbk`/`snapper` harness for empty and valid snapshot cases.

jameslai-dev:
What do you think about keeping the original `jq` query unmodified and skipping the iteration if `max_num` is `null`?
For example (based on your patch):

```diff
diff --git a/scripts/snapper-sync b/scripts/snapper-sync
index b49d07be..52613918 100755
--- a/scripts/snapper-sync
+++ b/scripts/snapper-sync
@@ -25,7 +25,7 @@ config_mapping=$(
 
 if outputs=$(snbk --machine-readable json list); then
   snapshots=$(
-    echo $outputs |
+    echo "$outputs" |
     jq --argjson config_mapping "$config_mapping" '
       .snapshots | . |= map(. + {config: $config_mapping[.name]})
     '
@@ -47,6 +47,14 @@ echo "$config_mapping" | jq -r 'keys[]' | while read -r backup_config; do
     jq -r "[.[] | select(.config == \"$source_config\")] | max_by(.number) | .number"
   )
 
+  # Nothing to synchronize for this source-config (e.g. no snapshots have been transferred
+  # and the source has none either). Skip to avoid creating a literal '.snapshots/null'
+  # placeholder directory.
+  if [ "$max_num" = "null" ]; then
+    echo "No snapshots found for source-config '$source_config'; skipping."
+    continue
+  fi
+
   # Query the source subvolume path
   subvol_path=$(
     snapper --machine-readable json -c "$source_config" get-config | jq -r '.SUBVOLUME'
```

This is sufficient to fix the problem while keeping changes to a minimum.

officialasishkumar:
Done in 05a79e2: kept the original `jq` query and now skip when `max_num` is `null`. I also reran the shell syntax check and the empty/valid snapshot harness.

aschnell:
Thanks for noticing the problem and providing a fix.

Issue #1011, opened 2025-04-30
How do I rollback when default subvolume is unknown?


rain1:
I installed Debian 12 so that root file system itself is on BTRFS to see if I have more luck with rollback that way.

This is how my root is mounted in ` cat /etc/fstab`:

```
# /etc/fstab: static file system information.
#
# Use 'blkid' to print the universally unique identifier for a
# device; this may be used with UUID= as a more robust way to name devices
# that works even if disks are added and removed. See fstab(5).
#
# systemd generates mount units based on this file, see systemd.mount(5).
# Please run 'systemctl daemon-reload' after making changes here.
#
# <file system> <mount point>   <type>  <options>       <dump>  <pass>
# / was on /dev/sda1 during installation
UUID=dbb22453-c02f-473c-b686-919c14924898 /               btrfs   defaults,subvol=@rootfs 0       0
# swap was on /dev/sda5 during installation
UUID=a4355a78-27bf-4058-8393-70c12a685d70 none            swap    sw              0       0
/dev/sr0        /media/cdrom0   udf,iso9660 user,noauto     0       0
```

Next I created config with command `snapper -c myconfig create-config /`

This is the output of command `snapper -c myconfig list`
```
 # | Type   | Pre # | Date                             | User | Cleanup | Description | Userdata
---+--------+-------+----------------------------------+------+---------+-------------+---------
0  | single |       |                                  | root |         | current     |
1  | single |       | Thu 01 May 2025 02:08:48 AM EEST | root |         |             |
```

But when I try to rollback with command `snapper -c myconfig rollback 1` I get:
```
Cannot detect ambit since default subvolume is unknown.
This can happen if the system was not set up for rollback.
The ambit can be specified manually using the --ambit option.
```

Do I need to recreate config or mount with different options?

aschnell:
With the option ```subvol=@rootfs``` in fstab I do not know how rollback is supposed to work. My knowledge about Debian is not enough to help here. I suggest you check Debian specific tutorials.

Issue #722, opened 2022-05-12
GRUB still looks for grub.cfg on previous subvolume after snapper rollback


geckolinux:
Hi there, on distros that are not based on openSUSE I keep running into an issue where the GRUB menu stops updating after a `snapper --ambit classic rollback` because GRUB is apparently looking for the `/boot/grub/grub.cfg` on the previous subvolume where `grub-install /dev/XXX` was performed. The workaround is to run `grub-install` again after rebooting into the new default subvolume that Snapper set. But this isn't ideal for end-users. Is there a distro-agnostic way of working around this? Or is there some fundamental concept about Btrfs and/or Snapper that I'm ignoring? Thanks for the help.

aschnell:
On openSUSE the package grub2-snapper-plugin provides the script /usr/lib/snapper/plugins/grub (https://build.opensuse.org/package/view_file/openSUSE:Factory/grub2/grub2-snapper-plugin.sh) which is called from snapper. But I cannot say if that is all that is required.

geckolinux:
Hi @aschnell , thanks for the reply. Yes, other distros don't include the GRUB2 snapper plugin. Even in openSUSE I'm not sure about its behavior, does Snapper on openSUSE somehow avoid this issue where the GRUB looks for the `grub.cfg` on the wrong (previous) subvolume where it was first installed? Is this behavior that I'm seeing expected, or am I not understanding something?

arashsm79:
I think openSUSE uses a patched version of grub that may not be available on all distros.
[Reliable Desktop Linux Installation with Btrfs Snapshots, Snapper, and Rollbacks on Ubuntu (openSUSE style)](https://moslehian.com/posts/2023/2-reliable-btrfs-snapshots-snapper-debian-ubuntu/) explains this issue and devises a snapper plugin that allows grub to look for `grub.cfg`  in the correct subvolume after each rollback.

#365 may also be relevant.

ak-li:
Possibly of interest.
[siduction-btrfs](https://github.com/siduction/siduction-btrfs)

We have developed a user-friendly solution to this problem.
siduction is a rolling release distribution based on Debian sid.

We use systemd units and two scripts to create a clean boot menu for the user.
In the case of GRUB, after a rollback, the system boots into the rollback target and reinstalls GRUB from there.
In the case of systemd-boot, boot entries are generated for the rollback target and all kernels present there.

Issue #159, opened 2015-03-13
Can't snapshot anymore after a "snapper rollback"


moy:
Hi,

I messed-up my system, and did a "snapper rollback <number>" and rebooted to get back to a consistant state. My system is repaired, but now, my default subvolume is not the root anymore, and snapper stops working.

```
$ sudo snapper --verbose -c root list  
Type   | # | Pre # | Date | User | Cleanup | Description | Userdata
-------+---+-------+------+------+---------+-------------+---------
single | 0 |       |      | root |         | current     |         
$ sudo snapper -c root create          
IO Error.
$ sudo tail -4 /var/log/snapper.log
2015-03-13 18:41:51 ERR libsnapper(22492) Snapshot.cc(initialize):366 - reading failed
2015-03-13 18:41:51 ERR libsnapper(22492) Btrfs.cc(openInfosDir):213 - .snapshots is not a btrfs snapshot
2015-03-13 18:42:04 ERR libsnapper(22492) Btrfs.cc(openInfosDir):213 - .snapshots is not a btrfs snapshot
2015-03-13 18:42:33 ERR libsnapper(22492) Btrfs.cc(openInfosDir):213 - .snapshots is not a btrfs snapshot
$ sudo ls /.snapshots/
$ 
```

The subvolumes are still there, but not visible in /.snaphots since I'm now using a different root:

```
$ sudo btrfs subvolume list -a /
ID 268 gen 3943 top level 5 path <FS_TREE>/.snapshots
ID 269 gen 1766 top level 268 path <FS_TREE>/.snapshots/1/snapshot
ID 273 gen 1806 top level 268 path <FS_TREE>/.snapshots/5/snapshot
ID 274 gen 1808 top level 268 path <FS_TREE>/.snapshots/6/snapshot
...
```

I'm not sure what's the expected behavior here (nor how I'm supposed to get back to a system that both works and is able to continue snapshoting), but either what I'm seeing is a bug, or at least the documentation of "snapper rollback" should be updated to explain the user that one of the consequences of doing a "snapper rollback" is to set the system to boot into a subvolume that doesn't show other snapshots and can't do snapshots anymore.

```
$ snapper --version
snapper 0.2.4
flags btrfs,lvm,ext4,xattrs,rollback,btrfs-quota
```

moy:
OK, I think I understand what happens: one needs to have a dedicated entry for /.snapshots/ in /etc/fstab like

```
[...]

PR #1133, opened 2026-05-15
rollback: support named-subvolume btrfs layouts (--ambit subvol-rename)

### LINKED ISSUE #365: Snapper rollback on Ubuntu does not use earlier snapshot
I am using a Ubuntu 16.04 environment

I recently performed the following:
a.) Took a snapper snapshot
b.) Performed an release upgrade
c.) Did a snapper rollback to pre-upgrade snapshot(basically go back to 16.04)
d.) reboot

The system did not boot with snapshot taken during a.)

Is this because the /boot/grub.d/grub.cfg in this case over-rides the work done by snapper rollback?

mxsb:
On systems where fstab mounts the root filesystem with an explicit `subvol=` option (e.g. `subvol=@root`), `snapper rollback` currently has no effect on the next boot. The `BTRFS_IOC_DEFAULT_SUBVOL` ioctl changes the default subvolume ID, but when the kernel mounts by name the default ID is ignored.

This affects systems using named-subvolume layouts (Fedora, Ubuntu, many manual btrfs setups) and is the root cause behind #365 and #1011 (and is discussed in related btrfs-rollback reports #722 and #159).

## Changes

The ambit is the single parameter controlling how the rollback is performed. `--ambit` accepts a fourth value:

- `auto` (default) — everything is detected: a top-level named `subvol=` root mount selects `subvol-rename`; otherwise the read-only/-write state of the default snapshot selects `transactional`/`classic`, exactly as before. No configuration needed on most systems.
- `classic` / `transactional` — existing behavior, unchanged (set-default).
- `subvol-rename` — atomically swaps the named root subvolume using `renameat2(RENAME_EXCHANGE)` (Linux 3.17+).

An explicit `--ambit` always wins over the detection. The whole decision lives in one pure function, `determine_ambit()` (`client/snapper/ambit.cc`), so the precedence is unit-tested. When an explicit `classic`/`transactional` sets the default subvolume ID while root is mounted by a top-level `subvol=` name — a rollback that cannot take effect on the next boot — snapper prints a warning.

`rollbackSubvolRename` is a virtual on `Filesystem` like `setDefault`, implemented for btrfs.

The `subvol-rename` sequence:
1. Read-write copy of the rollback target created as `@root.incoming`
2. `renameat2(RENAME_EXCHANGE)` atomically swaps `@root.incoming` ↔ `@root` — no window where `@root` does not exist
3. A nested `.snapshots` subvolume is moved into the new root (a btrfs snapshot only contains an empty stub directory where a nested subvolume is; without this the default Fedora/Debian layout would lose its snapshot history to the old root)
4. Old root moved to `@root.rollback.<N>` for recovery

Only a top-level named subvolume selects `subvol-rename`: the kernel shows the resolved `subvol=` path in `/proc/mounts` even for mounts by default subvolume ID (e.g. the nested `subvol=/@/.snapshots/N/snapshot` on Tumbleweed), so a nested name is no evidence of a by-name mount and such systems keep using set-default.

## Safety properties

- **Nested subvolume paths** are never selected by detection and are rejected with a clear error when requested explicitly, before any disk operations.
- **Old root is never deleted.** After the atomic swap, the old root is renamed to `@root.rollback.<N>`. If that name already exists, a subvolume-ID-based fallback name `@root.rollback.svid.<id>` is used — guaranteed unique across the btrfs filesystem.
- **Stale `.incoming` from a prior interrupted rollback** is renamed (not deleted) before starting. If a previous rollback completed the swap but failed to rename `.incoming`, that subvolume is the old running root, still mounted by the kernel via subvolume ID. Deleting it risks corruption during the unmount sequence. It is renamed to `@root.rollback.svid.<id>` instead.
- **`rename_exchange` failure** deletes only the newly-created `.incoming` copy (safe — the swap never occurred) and throws a clear error.

## Testing

- Unit tests for the ambit resolution (full precedence table incl. explicit-wins cases), the set-default-ineffective warning predicate, mount-option parsing, and the enum/name mapping (`testsuite/ambit-test.cc`)
- Unit tests for `SDir::exchange` / `RENAME_EXCHANGE` and the cross-directory `SDir::rename` (`testsuite/rename-exchange.cc`)
- Integration tests on real btrfs (`testsuite-real/rollback-subvol-rename.cc`): happy path, `.rollback.N` collision with subvolid fallback, nested-`.snapshots` migration semantics
- End-to-end VM tests on Fedora 43, Debian 13 and Tumbleweed (unattended install, build, rollback, reboot, verify — including a second rollback cycle with a forced `.rollback.N` collision, the `--ambit` conflict warning, and a transactional-ambit regression check): https://github.com/mxsb/snapper/tree/mxsb/testsuite-vm/testsuite-vm

## Documentation

- `snapper(8)` — `--ambit` values and the rollback section
- `doc/rollback.txt` — design rationale and mechanism descriptions

Fixes #365. Related: #722, #159, #1011.

---

> **Note:** This patch was developed with the assistance of Claude (Anthropic AI). The design, testing, and review decisions were made by the author.


mxsb:
I see a check failed. I’m currently on vacation, but will look into it asap when I’m back home.

aschnell:
The failure in Leap is unrelated to the changes.

snapper/AppUtil.h
```diff
@@ -83,6 +83,35 @@ namespace snapper
     bool getMtabData(const string& mount_point, bool& found, MtabData& mtab_data);
 
 
+#ifdef ENABLE_ROLLBACK
+
+    enum class RollbackMethod { SET_DEFAULT, SUBVOL_RENAME };
```
aschnell: Since the enum and the functions are only used in cmd-rollback.cc they should not be defined in the library.

client/snapper/cmd-rollback.cc
```diff
@@ -218,7 +273,20 @@ namespace snapper
...
-		filesystem->setDefault(snapshot2->getNum(), report);
+		switch (rollback_method)
+		{
+		    case RollbackMethod::SET_DEFAULT:
+			filesystem->setDefault(snapshot2->getNum(), report);
+			break;
+		    case RollbackMethod::SUBVOL_RENAME:
+		    {
+			const Btrfs* btrfs = dynamic_cast<const Btrfs*>(filesystem.get());
+			if (!btrfs)
```
aschnell: There is already a check for btrfs above. So the variable btrfs could also be defined there.

Another possibility would be to add rollbackSubvolRename to Filesystem like it is done for setDefault. Esp. since renaming subvolume could maybe also be used to rollback using LVM (there the logical volumes would be renamed). Although currently there are no plans to implement this for LVM.

aschnell:
I installed Fedora 44 and the mount option in fstab is subvol=root (not subvol=@root). The rollback fails.
It also cannot detect the ambit. Do I need special settings during installation of Fedora?

aschnell:
Now I installed Debian 13.5 and get the same error. Although in /etc/fstab subvol=@rootfs is used in /proc/mounts it's subvol=/@rootfs. So the check again fails.

snapper/FileUtils.cc
```diff
@@ -400,6 +401,28 @@ namespace snapper
...
+	assert(name2 != "..");
+
+#ifndef RENAME_EXCHANGE
+#define RENAME_EXCHANGE (1 << 1)
+#endif
+
+#ifdef HAVE_RENAMEAT2
+	return ::renameat2(dirfd, name1.c_str(), dirfd, name2.c_str(), RENAME_EXCHANGE);
+#else
+	return syscall(SYS_renameat2, dirfd, name1.c_str(), dirfd, name2.c_str(),
```
aschnell: ```renameat2``` was added in glibc in 2018 so this fallback looks unnecessary. Even without the workaround it builds on all archs currently tested (e.g. https://build.opensuse.org/project/show/home:aschnell:snapper).

mxsb:
Thanks for your comments and looking at the PR, I'm working on the feedback.

snapper/AppUtil.h
mxsb: fixed

client/snapper/cmd-rollback.cc
mxsb: fixed

snapper/FileUtils.cc
mxsb: fixed

mxsb:
I reworked the whole PR and hope it's in better shape. Also fixed a few bugs, e.g. the ambit detection.
I created some local qemu test images to test this with Debian, Fedora and Tumbleweed. Maybe I polish it a bit, so it can be used in this repo, are you interested?

Thanks again for your comments and tests, let me know if you feel something should be reworked.

mxsb:
Hey @aschnell , 
I'd appreciate if you'd have the chance to look at this PR again. You can check my testing methodology here https://github.com/mxsb/snapper/commit/5a4ef4c47e4b5cc36345f3538997b65727c0dfaa - its testing Fedora 43, Tumbleweed and Debian 13. Happy to add it to the PR and contribute, just didn't want to increase the scope and was not sure if you find it useful. 

Let me know what you think and if you feel that rework is required, I'm happy to adjust the PR.

aschnell:
I was out of office and didn't have time yet to look at it again.

aschnell:
Can the new RollbackMethod enum be avoided by adding a new value to the existing Ambit enum? The current ambit only controls the rollback so that looks reasonable to me.

snapper/AppUtil.cc
```diff
@@ -41,6 +43,7 @@
 
 #include "snapper/LoggerImpl.h"
 #include "snapper/AppUtil.h"
+#include "snapper/Exception.h"
```
aschnell: Unneeded change.

snapper/AppUtil.cc
```diff
@@ -21,6 +21,8 @@
  */
 
 
+#include "config.h"
```
aschnell: Unneeded change.

snapper/FileUtils.cc
```diff
@@ -25,6 +25,7 @@
 
 #include <cstdio>
 #include <cstring>
+#include <sys/syscall.h>
```
aschnell: Unneeded change.

client/snapper/GlobalOptions.h
```diff
@@ -40,7 +43,6 @@ namespace snapper
     public:
 
 	enum class OutputFormat { TABLE, CSV, JSON };
-	enum class Ambit { AUTO, CLASSIC, TRANSACTIONAL };
```
aschnell, line 43: No need to move Ambit somewhere else. If you don't want to write ```GlobalOptions::Ambit``` so often somewhere you can use ```using Ambit = GlobalOptions::Ambit;```

client/snapper/GlobalOptions.h
```diff
@@ -24,7 +24,10 @@
 
 #include <string>
 
+#include <snapper/AppUtil.h>
 #include <snapper/Enum.h>
+
+#include "rollback-method.h"
```
aschnell: When the Ambit enum stays here the include statement is not needed. Another reason to keep the enum here.

client/snapper/GlobalOptions.h
mxsb: Done — `GlobalOptions` is unchanged from master again; `Ambit` stays as `GlobalOptions::Ambit`, and `rollback-method.h` uses `using Ambit = GlobalOptions::Ambit;` as you suggested.

client/snapper/GlobalOptions.h
mxsb: Done — `GlobalOptions.h` no longer includes `rollback-method.h`; the dependency is reversed (`rollback-method.h` includes `GlobalOptions.h`), so `GlobalOptions.h` is back to its original state.

snapper/AppUtil.cc
mxsb: Reverted.

snapper/AppUtil.cc
mxsb: Reverted.

snapper/FileUtils.cc
mxsb: Reverted — it was left over from the removed `renameat2` syscall fallback.

mxsb:
Good point @aschnell — the `RollbackMethod` enum is gone.

I first tried folding it into `Ambit` as you suggested, but adding a `SUBVOL_RENAME` value mixed the rollback *mechanism* (set-default vs subvolume-rename) into the *ambit*, which otherwise only distinguishes classic from transactional: the switch then needed a shared `case CLASSIC: case SUBVOL_RENAME:` with an inner `if`, and `Ambit` lost its single meaning.

So `Ambit` stays `{ AUTO, CLASSIC, TRANSACTIONAL }` and the mechanism is a predicate rather than an enum:

- `use_subvol_rename(rollback_method, subvol_name)` → bool, from `ROLLBACK_METHOD` and the root mount
- `resolve_ambit(ambit, mode)` → the classic/transactional ambit

The rollback switches on the ambit only; within the classic case the bool picks `renameat2` over set-default. Both helpers are pure and unit-tested. This removes the extra enum as you asked while keeping the two concerns separate — but if you'd rather have the `SUBVOL_RENAME` value on `Ambit`, I have that variant ready too and can switch.

client/snapper/cmd-rollback.cc
```diff
@@ -130,31 +137,54 @@ namespace snapper
...
 	ProxySnapshots::iterator previous_default = snapshots.getDefault();
 
-	if (global_options.ambit() == GlobalOptions::Ambit::AUTO)
+	const string subvol_name = get_subvol_name(subvolume);
+
+	string rollback_method;
+	config.getValue(KEY_ROLLBACK_METHOD, rollback_method);
+
+	const bool subvol_rename = use_subvol_rename(rollback_method, subvol_name);
+
```
aschnell, line 141: Maybe I'm wrong but the code below completely ignores the ambit as provides by the user.

aschnell:
I still think it's confusing to have ambit and rollback-method. In the end it controls how the rollback
is performed and having only one parameter should be more understandable. If it is required to
control it in the configuration file the ambit can be added there.

mxsb:
Thank you for your review! The ambit is now the single parameter. `ROLLBACK_METHOD` is gone entirely; instead `--ambit` accepts a fourth value `subvol-rename`, and with the default `auto` everything is detected: a top-level named `subvol=` root mount selects subvol-rename, otherwise the read-only/-write state of the default snapshot selects transactional/classic, exactly as before. An explicit `--ambit` always wins over the detection. I did not add a config key for now since auto covers the named-subvolume distros without configuration — an `AMBIT` key can be added later if you see a need.

Two more changes from testing this against real Fedora 43 / Debian 13 / Tumbleweed VMs:

- `rollbackSubvolRename` is now a virtual on `Filesystem` like `setDefault` (your earlier suggestion), so the `dynamic_cast` in cmd-rollback.cc is gone.
- The rollback now moves a nested `.snapshots` subvolume into the new root after the swap. A btrfs snapshot only contains an empty stub directory where a nested subvolume is, so without this the default Fedora/Debian layout (`.snapshots` inside the root subvolume) lost its snapshot history to the renamed-away old root — caught by a repeated-rollback VM test.

Since `RollbackMethod`/`ROLLBACK_METHOD` no longer exists, `rollback-method.{h,cc}` is renamed to `ambit.{h,cc}`.

client/snapper/cmd-rollback.cc
mxsb: You were right — fixed. The whole decision now lives in one pure function, `determine_ambit(cli_ambit, subvol_name, mode)`: an explicit `--ambit` always wins, detection only fills in `auto`. This is unit-tested including the cases you found (`--ambit classic`/`transactional` on a named-subvolume mount must not be overridden by detection). Since an explicit `--ambit classic` on such a mount produces a rollback that has no effect on the next boot, snapper now prints a warning for that combination — verified end-to-end in the VM tests on Fedora and Debian.

snapper/Btrfs.cc
```diff
@@ -1513,6 +1513,148 @@ namespace snapper
     }
 
 
+    void
+    Btrfs::rollbackSubvolRename(unsigned int num, const string& subvol_name,
```
aschnell, line 1517: Btrfs::rollbackSubvolRename must also be implemented in the other branch of the ENABLE_ROLLBACK if. Just like e.g. isDefault. Otherwise building fails if rollback is disabled.

snapper/FileUtils.h
```diff
@@ -97,6 +97,7 @@ namespace snapper
 	int chmod(const string& name, mode_t mode, int flags) const;
 	int chown(const string& name, uid_t owner, gid_t group, int flags) const;
 	int rename(const string& oldname, const string& newname, int flags = 0) const;
+	int rename(const string& oldname, const SDir& new_dir, const string& newname) const;
```
aschnell, line 100: Please also add ```int flags = 0``` here.

client/snapper/ambit.h
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Whether a rollback with the given ambit sets the default subvolume id
+     * while root is mounted with a top-level subvol= name (subvol_name) - the
+     * kernel then ignores the default subvolume id and the rollback has no
+     * effect on the next boot. Used to warn when an explicit --ambit conflicts
+     * with the root mount. False for a nested name, which the kernel also
+     * shows for a mount by the default subvolume id.
+     */
+    bool set_default_ineffective(Ambit ambit, const string& subvol_name);
```
aschnell, line 76: Should better be named ```is_set_default_ineffective```.

aschnell:
I just tested on Fedora 44. The ambit is detected correctly and rollback happens. But before rebooting
/.snapshots does not exist so snapper does not work anymore. After reboot it's back though. Maybe
/.snapshots must be listed in fstab.

Another problem I see are the subvolumes. E.g. I have root.incoming.13 and also root.incoming.13/var/lib/portables.
The problem with the first is that AFAIS it will never be deleted by snapper. The second looks like
the result of some misconfiguration.

aschnell:
Not in a mergeable state.

mxsb:
Since this PR is closed, I opted to open a clean new PR — #1184 — that continues the work here with all of your feedback addressed:

- `/.snapshots` unavailable before reboot — the nested `.snapshots` is now re-mounted onto the running root by its subvolume id after the move, so snapper keeps working immediately (transient; the new root provides it again after reboot).
- `.incoming` left behind / backups never deleted — `.incoming` is renamed to `<subvol>.rollback.<N>` after the swap, and a new optional `ROLLBACK_BACKUP_LIMIT` prunes old backups (empty = keep all, the default).
- nested `var/lib/portables` in the copy — expected btrfs behaviour (a snapshot doesn't recurse into nested subvolumes); only `.snapshots` needs moving, the rest are empty stubs removed when the backup is pruned.

The VM test suite has also grown to Fedora 43, Debian 13, Ubuntu 26.04 and Tumbleweed (all green) and is offered as a separate contribution. Details in #1184.

PR #1150, opened 2026-06-22
Fix ssh option configuration ordering for snbk ssh-push targets

Devansh-567:
## Description
This PR fixes an issue within the `snbk` client engine where configured SSH parameters (such as custom ports, identity files, user logins, and control multiplexing options) are mistakenly evaluated as remote commands instead of local execution flags.

---

## Problem
In `client/snbk/BackupConfig.cc`, the `ssh_options()` method constructs the argument vector starting with the target host variable as its very first element:

```cpp
vector<string> options = { ssh_host };
```

When client/snbk/Shell.cc subsequently runs the shell runtime array via ssh <options> <command>, it appends the target command block onto this options array. According to standard OpenSSH syntax rules, any flags positioned after the hostname parameter are interpreted as part of the remote terminal command string rather than local client configurations.

Consequently, parameters like -p, -l, -i, and -o fail to apply to the local SSH session, breaking custom connectivity rules entirely.

## Solution
Modified BackupConfig::ssh_options() to initialize an empty vector, sequentially inject the connection flags first, and append ssh_host as the final parameter of the configuration block. This satisfies standard ssh(1) argument sequencing rules.

## Bug Discovery & Technical Audit Analysis
- Context
While reviewing the snbk push-target ecosystem to map out the execution pipeline of backup synchronization sequences, a manual engineering code inspection was performed across the configuration setup modules (client/snbk/BackupConfig.cc) and shell execution wrapper scopes (client/snbk/Shell.cc).

1. Tracing the Execution Flow: In client/snbk/Shell.cc, the engine compiles and executes an active shell runtime command array structured identically to a standard ssh <options> <command> interface pattern.
2. Identifying the Structural Anomaly: Looking closely at the argument builder inside BackupConfig::ssh_options(), the configuration array was initialized with the target host tracking variable placed at index 0, followed by a sequence of conditional .insert() operations.
3. Root Cause Evaluation: In POSIX and OpenSSH CLI specifications, flags governing transport parameters (such as routing ports via -p, identity credentials via -i, or login aliases via -l) must be parsed prior to parsing the remote host target. Because the initial implementation positioned ssh_host at the absolute front of the vector array, any runtime configuration flag appended subsequently was pushed down into the remote command payload block, completely bypassing the local client configuration parsing sequence.

This configuration sequencing error was caught through structural source code analysis during a proactively scheduled code health audit. The patch cleanly rearranges the vector assembly mechanics to align with proper argument positioning constraints.

### DIFF
--- a/client/snbk/BackupConfig.cc
+++ b/client/snbk/BackupConfig.cc
@@ -107,7 +107,7 @@ namespace snapper
     vector<string>
     BackupConfig::ssh_options() const
     {
-	vector<string> options = { ssh_host };
+	vector<string> options;
 
 	if (ssh_port != 0)
 	    options.insert(options.end(), { "-p", to_string(ssh_port) });
@@ -121,6 +121,9 @@ namespace snapper
 	if (ssh_master_control)
 	    options.insert(options.end(), { "-o", "ControlMaster=auto", "-o", "ControlPath=\"~/.ssh/snbk-%C\"",
 		"-o", "ControlPersist=30s" });
+
+	// The target host must be the final argument before the command string execution block
+	options.push_back(ssh_host);
 
 	return options;
     }


aschnell:
Thanks for the review and patch.

Is there a way to force ssh to respect the strict ordering for testing? AFAIS POSIXLY_CORRECT does not do so.

Devansh-567:
> Thanks for the review and patch.
> 
> Is there a way to force ssh to respect the strict ordering for testing? AFAIS POSIXLY_CORRECT does not do so.

Hi @aschnell,

Thank you for merging the patch! 

Regarding strict ordering enforcement for validation/testing, you are exactly right, `POSIXLY_CORRECT` has no effect here because OpenSSH uses its own custom command-line argument parser rather than standard GNU `getopt`.

However, we can force OpenSSH to evaluate or expose strict argument sequencing issues using two distinct approaches:

### 1. Forcing execution failures via remote command interpretation
Since OpenSSH treats any flag positioned after the host string as part of the remote command string, we can inject a target command that catches unexpected argument arrays. 

For example, if `snbk` tries to execute a simple remote echo command (`echo "syncing"`), the out-of-order configuration string transforms into:
```bash
ssh user@host -p 2222 echo "syncing"
```
Because -p 2222 falls inside the command execution block, the remote shell literally attempts to execute a command named -p. This results in an immediate and predictable stderr termination:
bash: line 1: -p: command not found

In an automated integration test, checking if stderr captures a command-not-found diagnostic on option flags (like -p, -l, -i) effectively flags a regression.

### 2. Validating arguments directly using SSH's dry-run flag (-G)
If you want to explicitly programmatically parse what local parameters the client thinks it is using without spawning an active remote connection, OpenSSH supports the -G (Print Configuration) flag.

When passed before the host, it causes SSH to print the fully parsed local configuration mapping to stdout and exit:

```bash
# Correct evaluation order (Prints full routing block details to stdout)
ssh -p 2222 -G user@host
```
If the arguments are incorrectly sequenced (as they were previously):
```bash
ssh user@host -p 2222 -G
```

The client skips local evaluation of -p 2222 -G entirely and passes them over to the remote host pipeline, preventing any configuration layout from printing locally. A test assertion checking that stdout from a -G validation pipeline is not empty provides an accurate diagnostic tool.

PR #1151, opened 2026-06-22
Decouple DBus pipe file descriptors via dup() to prevent double-close bugs

Devansh-567:
## Description
This PR addresses a critical resource management bug where file descriptors are tracked under dual ownership models inside both `server/FilesTransferTask.cc` and `client/proxy/commands.cc`, causing double-close race conditions.

---

## Problem Analysis
When passing a file descriptor to `fdopen()`, the resulting `FILE*` stream takes full ownership of that descriptor. When `fclose()` is subsequently invoked, the underlying OS descriptor is closed.

Currently, the codebase invokes `fdopen()` directly on descriptor integers managed by an RAII wrapper (`DBus::FileDescriptor`):
```cpp
FILE* fin = fdopen(fd.get_fd(), "r");
```
## Solution
Introduced dup() when passing descriptors into fdopen(). This clones the descriptor within the kernel table, separating the resource lifecycles:
- fclose() safely tears down the duplicated descriptor.
- The DBus::FileDescriptor destructor safely tears down the original descriptor.
Both locations have also been updated to include <unistd.h>.

### Context
Following up on stability validation tasks within the cross-process IPC communication boundaries, a structural manual engineering inspection was conducted on POSIX stream interfaces across the transfer tasks (server/FilesTransferTask.cc) and proxy command modules (client/proxy/commands.cc).

### 3. Root Cause Evaluation: Audited the FileDescriptor class definitions inside dbus/DBusPipe.h. The class lacks an explicit ownership transfer or release layout (.release() or .take()). As a result, standard usage inevitably patterns into overlapping closure scopes when interacting with standard library C-streams.

This issue was successfully isolated via manual static architectural analysis during a proactively scheduled code health quality sweep.

### DIFF
--- a/client/proxy/commands.cc
+++ b/client/proxy/commands.cc
@@ -554,7 +554,7 @@ command_get_xfiles_by_pipe(DBus::Connection& conn, const string& config_name, un
 
     vector<XFile> files;
 
-    FILE* fin = fdopen(fd.get_fd(), "r");
+    FILE* fin = fd.fdopen("r");
     if (!fin)
 	SN_THROW(IOErrorException("reading pipe failed, fdopen failed: " + stringerror(errno)));
 
--- a/dbus/DBusPipe.cc
+++ b/dbus/DBusPipe.cc
@@ -65,6 +65,14 @@ namespace DBus
 	return marshaller;
     }
 
+    FILE*
+    FileDescriptor::fdopen(const char* mode)
+    {
+        FILE* ret = ::fdopen(_fd, mode);
+        if (ret)
+            _fd = -1;
+        return ret;
+    }
 
     Pipe::Pipe()
     {
--- a/dbus/DBusPipe.h
+++ b/dbus/DBusPipe.h
@@ -44,6 +44,8 @@ namespace DBus
 	int get_fd() const { return _fd; }
 	void set_fd(int fd) { _fd = fd; }
 
+	FILE* fdopen(const char* mode);
+
     private:
 
 	int _fd = -1;
--- a/server/FilesTransferTask.cc
+++ b/server/FilesTransferTask.cc
@@ -33,7 +33,7 @@ FilesTransferTask::FilesTransferTask(const Files& files)
 void
 FilesTransferTask::run()
 {
-    FILE* fout = fdopen(get_write_end().get_fd(), "w");
+    FILE* fout = get_write_end().fdopen("w");
     if (!fout)
 	SN_THROW(StreamException());
 


aschnell:
The analysis seems correct. But I think adding a release_fd() for fdopen() function to DBus::FileDescriptor looks
like a better fix.

Devansh-567:
> The analysis seems correct. But I think adding a release_fd() for fdopen() function to DBus::FileDescriptor looks like a better fix.

Hi @aschnell,

You are completely right. Adding an explicit ownership-transfer mechanism directly to our RAII wrapper is a much more elegant architectural fix than forcing a kernel-level table duplication via `dup()`.

I have updated the patch to:
1. Implement an explicit `release_fd()` utility method inside `DBus::FileDescriptor` which clears the internal tracking reference to `-1` and returns the native descriptor index.
2. Swapped the logic blocks inside `client/proxy/commands.cc` and `server/FilesTransferTask.cc` to feed `fdopen()` directly with the detached descriptors, passing clean asset ownership off to `fclose()`.

The branch has been updated and force-pushed. Let me know if this implementation looks good to you now!

aschnell:
The release_fd has the disadvantage that if fdopen fails (although very unlikely) the fd is never closed.

An fdopen function in DBus::FileDescriptor should be safer:

```
FILE*
FileDescriptor::fdopen(const char* mode)
{
    FILE* ret = ::fdopen(_fd, mode);
    if (ret)
        _fd = -1;
    return ret;
}
```

Devansh-567:
> The release_fd has the disadvantage that if fdopen fails (although very unlikely) the fd is never closed.
> 
> An fdopen function in DBus::FileDescriptor should be safer:
> 
> ```
> FILE*
> FileDescriptor::fdopen(const char* mode)
> {
>     FILE* ret = ::fdopen(_fd, mode);
>     if (ret)
>         _fd = -1;
>     return ret;
> }
> ```

Hi @aschnell,

That is an incredibly sharp catch! You're completely right, if `::fdopen` failed after calling `release_fd()`, the descriptor would completely leak since the RAII wrapper would have already disowned it. 

Encapsulating the logic directly inside a member `fdopen` function provides defensive safety by only shifting ownership (`_fd = -1`) upon a confirmed successful stream allocation.

I have refactored the patch to:
1. Remove `release_fd()` and instead declare/define `FILE* FileDescriptor::fdopen(const char* mode)` within `dbus/DBusPipe.h` and `dbus/DBusPipe.cc` using the global namespace resolution operator `::fdopen`.
2. Updated the call sites in `client/proxy/commands.cc` and `server/FilesTransferTask.cc` to use the safe method wrapper.

The branch has been updated and force-pushed. Thank you for the great review!

aschnell:
Thanks, looks good now.

PR #1155, opened 2026-06-22
Fix: Validate STOMP content-length header to prevent integer parsing crashes

Devansh-567:
### Description
This PR hardens the STOMP frame parser against malformed or hostile network inputs. Specifically, it ensures that the `content-length` header is rigorously validated before its value is used to allocate memory for the frame body.

### Problem
In `stomp/Stomp.cc`, the parser extracts the `content-length` header string and passes it directly to `std::stol(value.c_str())`. While `std::stol` works fine for ideal inputs, it introduces three significant edge-case vulnerabilities when handling untrusted protocol data:

1. **Partial/Malformed Parsing:** `std::stol` stops parsing at the first non-numeric character. For instance, a header like `content-length: 100abc` is interpreted as `100`. This masks malformed frames that should otherwise fail fast as protocol errors.
2. **Uncaught Exceptions / DoS:** Passing a value that exceeds the maximum limit of a signed `long` causes `std::stol` to throw an unhandled `std::out_of_range` exception, abruptly crashing the process.
3. **Negative Bounds / Memory Underflow:** If a negative value is supplied (e.g., `content-length: -5`), `std::stol` successfully parses `-5`. Later, when initializing the buffer with `vector<char> buf(content_length);`, the signed `ssize_t` is implicitly cast to an unsigned `size_t`. This triggers a massive memory allocation attempt (close to `SIZE_MAX`), resulting in an immediate `std::bad_alloc` or runtime crash.

### How I Found This
While reviewing the `stomp` engine architecture and auditing how network-facing inputs boundary checks are handled, I noticed that `content_length` directly dictates the size of the allocation buffer `vector<char> buf(content_length)`. 

Tracing back to where `content_length` is populated, it relied entirely on `std::stol` without pre-validation. To verify the vulnerability, I tested the parser behavior with simulated edge-case inputs:
- An overflow value (`99999999999999999999`) caused an uncaught `out_of_range` exception.
- A negative value (`-10`) bypassed the `if (content_length > 0)` logic after the unsigned implicit cast during vector construction, leading to allocation failure.

This confirmed that the parser lacked strict string-to-integer sanitization common in robust network protocols.

### Solution
I have refactored the header evaluation block using standard `<algorithm>` and structured exception handling:

1. **Character Sanitization:** Added an explicit `std::all_of` check utilizing `::isdigit` to guarantee the string contains strictly numeric characters. This naturally eliminates empty strings, partial trailing characters (`10abc`), and embedded signs (`-` or `+`).
2. **Safe Parsing:** Upgraded the parsing to `std::stoll` wrapped inside a explicit `try-catch` block to gracefully capture `std::invalid_argument` and `std::out_of_range`.
3. **Explicit Boundary Checks:** Added an explicit sanity check to ensure the length is non-negative before narrowing the type and casting to `ssize_t`. 

Malformations now reliably bubble up as standard `runtime_error("stomp error: ...")` messages, allowing the upstream session handler to gracefully drop the malformed connection instead of crashing the entire application.

Additionaly if you find any cosmetic errors please do let me know :)

### DIFF
--- a/stomp/Stomp.cc
+++ b/stomp/Stomp.cc
@@ -24,7 +24,6 @@
 
 #include "Stomp.h"
 
-
 namespace Stomp
 {
 
@@ -106,7 +105,35 @@ namespace Stomp
 		    if (key == "content-length")
 		    {
 			has_content_length = true;
-			content_length = stol(value.c_str());
+
+			if (value.empty())
+			{
+			    throw runtime_error("stomp error: empty content-length value");
+			}
+
+			try
+			{
+			    size_t parsed_chars = 0;
+			    long long parsed_length = stoll(value, &parsed_chars);
+
+			    // 1. Check if there are trailing unparsed characters (e.g., "100abc")
+			    // 2. Reject negative integer limits
+			    // 3. Explicitly reject negative signs to enforce pure digits
+			    if (parsed_chars < value.size() || parsed_length < 0 || value[0] == '-')
+			    {
+				throw runtime_error("stomp error: invalid content-length value '" + value + "'");
+			    }
+
+			    content_length = static_cast<ssize_t>(parsed_length);
+			}
+			catch (const invalid_argument&)
+			{
+			    throw runtime_error("stomp error: invalid content-length syntax");
+			}
+			catch (const out_of_range&)
+			{
+			    throw runtime_error("stomp error: content-length value out of range");
+			}
 		    }
 
 		    msg.headers[key] = value;
@@ -116,7 +143,6 @@ namespace Stomp
 
 	throw runtime_error("stomp error: expected a message, got a part of it");
     }
-
 
     void
     write_message(ostream& os, const Message& msg)


stomp/Stomp.cc
```diff
@@ -106,7 +106,32 @@ namespace Stomp
 		    if (key == "content-length")
 		    {
 			has_content_length = true;
-			content_length = stol(value.c_str());
+
+			// Validate that the string consists solely of numeric digits
+			if (value.empty() || !std::all_of(value.begin(), value.end(), ::isdigit))
```
aschnell: This is not needed. Better check the str_end parameter of stoll.

Devansh-567: > This is not needed. Better check the str_end parameter of stoll.

That makes total sense! Leveraging `stoll`'s tracking pointer to catch unconverted trailing characters lets us handle this in a single pass without needing a separate validation loop. 

I've refactored the validation block to use the tracking index, explicitly enforced checks against negative inputs/signs, and removed the unnecessary `<algorithm>` include to keep the patch clean. Force-pushed the update. Thanks for the guidance!

aschnell:
Thanks.

PR #1162, opened 2026-06-25
Release parsed zypp JSON objects

Devansh-567:
## Description
This pull request fixes a memory leak in the snapper zypp plugin. The function responsible for extracting solvable package entries from incoming zypp message payloads parses JSON text into an in-memory object tree but fails to release the allocated memory before returning.

## Problem
In `zypp-plugin/snapper-zypp-plugin.cc`, the method `SnapperZyppCommitPlugin::get_solvables` handles JSON payload translation. It invokes `json_tokener_parse_ex` to build a root `json_object` node tree. 

While the surrounding `JsonTokener` class relies on its RAII destructor to cleanly free the underlying parsing mechanism wrapper, the generated `json_object` tree itself operates on reference counting. Because no execution paths ever trigger a balancing `json_object_put(zypp)` call, the parsed structures are leaked on every payload processing evaluation.

## Solution
Modified the control flow within `get_solvables` to safely handle lifecycle cleanups:
- Removed the raw early-exit return (`if (!steps) return result;`) that bypassed cleanup.
- Wrapped the array extraction logic inside a safe, nested conditional check (`if (steps)`), forcing execution to collapse to the baseline of the function under all runtime circumstances.
- Added a explicit `json_object_put(zypp);` statement immediately before returning the final package result set to release the heap allocation.

## How I Found It
While reviewing the plugin source file to analyze payload transaction tracking and memory profile footprint structures, I observed a structural discrepancy. The token parsing lifecycle implementation ensures structural parsing structures balance out, but the generation layer completely omitted the mandatory reference release step defined by the json-c memory usage interface guidelines.

### DIFF
--- a/zypp-plugin/snapper-zypp-plugin.cc
+++ b/zypp-plugin/snapper-zypp-plugin.cc
@@ -356,51 +356,52 @@ SnapperZyppCommitPlugin::get_solvables(const Message& msg, Phase phase) const
     // {"TransactionStepList":[{"type":"?","stage":"?","solvable":{"n":"mypackage"}}]}
     // https://doc.opensuse.org/projects/libzypp/SLE12SP2/plugin-commit.html
     json_object* steps = object_get(zypp, "TransactionStepList");
-    if (!steps)
-	return result;
-
-    if (json_object_get_type(steps) == json_type_array)
+    if (steps)
     {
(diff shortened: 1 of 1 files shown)


aschnell:
Thanks for the patch.

Issue #1105, opened 2026-03-05
SNBK/Snapper Backup Default Action/Retention Policy Discussion


trainzkid:
On [issue #1096](https://github.com/openSUSE/snapper/issues/1096#issuecomment-3888438809), I said: 
> Yeah I'm not super sure why the default action is to remove potentially usable snapshots

@jamesljlster responded with: 
> I assume the retention policy is designed to follow the source Snapper config (please correct me if I'm wrong, Arvin). Would you mind opening a separate issue to discuss this topic further?

This issue is to discuss further. The context is that I had old snapshots from before I moved to a new storage drive (HDD -> SSD) that I wanted to keep, as they were good snapshots taken following a snapper timeline event. `snbk` deletes snapshots it considers "invalid", so I had to move things around so the destination/target directory was "clean" so I could retain my old snapshots. 

Though the more I explain, the more it sounds like I should have just manually `btrfs send | btrfs receive`d them across so they were present in the source directory following the migration from old storage to new storage. Thoughts?

jamesljlster:
I believe the technical detail of how `snbk` determines an invalid backup is a professional design.
However, I'm wondering whether automatically removing invalid snapshots might be too aggressive. A real scenario could occur when an operator restores data without restoring the snapshots, and then `snapper` creates a snapshot with a number collision. In this case, `snbk` might determine the colliding snapshot invalid and remove it on the backup device, which may not be the appropriate action.

Because of this, I'm considering whether it would be better to let users decide how to handle invalid snapshots. During automatic backups, `snbk` could simply ignore invalid backups and continue processing the remaining snapshots.

By the way, what do you think about the retention policy design?

jamesljlster:
> I think removing invalid snapshots and then re-transfer the snapshot is the only way to go. Otherwise the backup might not include snapshots that should be included.

What a great insight. I might have been focusing too much on preserving the old data.

> One way the improve the situation would be to not reuse snapshots numbers. Then the only reason to get invalid snapshots should be an interrupted transfer.

This sounds reasonable. However, I wonder whether it is feasible in the current architecture. Please correct me if I'm wrong. Currently, `snapperd` is not aware of the backed-up snapshots created by `snbk`. This could make it difficult to detect the highest snapshot number across both the source and the target.
I think it may be more appropriate to discuss this topic in another issue. I have seen a few related issues already, so perhaps we could continue the discussion in one of those.

> About a different cleanup policy on the target than on the source: Yes, that would be possible.

[...]

PR #1167, opened 2026-07-13
Add snapshot metadata loading for snbk

jameslai-dev:
This is the initial PR for developing the decoupled snapshot retention policy for `snbk`, as discussed in #1105 (https://github.com/openSUSE/snapper/issues/1105#issuecomment-4457421134).
This PR implements snapshot metadata loading for `snbk`, which adds the necessary snapshot attributes for cleanup algorithms to `TheBigThing`: `type`, `pre_num`, `cleanup`, and `userdata`.

Detailed changes:
- The `TheBigThing::load_metadata()` method is added to load snapshot metadata and is executed after `probe_source` and `probe_target`.
  Since the backup device may be slower than the source device (e.g., via SSH), if the source snapshot is present, the metadata is copied from the source snapshot; otherwise, the metadata is loaded from the backup snapshot.
- Since `info.xml` is not directly accessible when using SSH push, a constructor is added to `XmlFile` to load the XML content from a string.


jameslai-dev:
Just realized I should use `-p` instead of `-P` for specifying the port number for `ssh`. It's now fixed. Sorry!

client/snbk/TheBigThing.cc
```diff
@@ -649,6 +653,105 @@ namespace snapper
...
+		else
+		{
+		    the_big_thing.type = tmp_type;
+		}
+
+		getChildValue(node, "pre_num", the_big_thing.pre_num);
+		getChildValue(node, "cleanup", the_big_thing.cleanup);
+
+		const vector<const xmlNode*> l = getChildNodes(node, "userdata");
+		for (vector<const xmlNode*>::const_iterator it2 = l.begin();
```
aschnell, line 741: Just use ```for (const xmlNode* tmp : getChildNodes(node, "userdata"))```.

aschnell:
Please also increase the minor version number in LIBVERSION.

client/snbk/TheBigThing.cc
```diff
@@ -649,6 +653,105 @@ namespace snapper
...
+		switch (backup_config.target_mode)
+		{
+		    case BackupConfig::TargetMode::LOCAL:
+		    {
+			cmd_args << CAT_BIN;
+		    }
+		    break;
+
+		    case BackupConfig::TargetMode::SSH_PUSH:
+		    {
```
aschnell, line 689: Why can't shellify be used here? Maybe also create a CmdCat class to handle the target mode and have simpler code here.

aschnell:
AFAIS with the change there are two calls for info.xml on the target if the snapshot is not on the source: Once sha256sum and once cat. That could be done in one (cat) by calculation the sha256sum in snbk (likely using libopenssl).

On the other hand the sha256sum is only needed if the snapshot also exists on the source. So that call could be avoided
if the snapshot only exists on the target. Then it would be either sha256sum or cat.

Since so far the list of snapshots is normally almost identical this optimization does not really matter.

client/snbk/TheBigThing.cc
jameslai-dev: It seems possible. Thanks for your advice!

jameslai-dev:
Thanks for your review. What about wrapping another class for `info.xml` that provides the file checksum and content?
The `info.xml` files only need to be read once. Checksum calculation and content loading can be handled entirely from memory.
Since the final goal is to decouple the retention policies of the source and target, the list of snapshots will not be identical in the future. I believe the optimization will be important.

aschnell:
Not sure how this wrapper for info.xml is supposed to look but in general it sound good.

jameslai-dev:
I implemented a `CmdMetaParse` in another branch:
https://github.com/openSUSE/snapper/compare/master...jameslai-dev:snapper:dev-snbk-cmd-meta-parse

The `CmdMetaParse` provides the checksum and metadata on demand. Overall, this implementation makes fewer changes to the `TheBigThing` module. What do you think?
If you prefer the new implementation, I can close this PR and open another one to submit the new patches.

aschnell:
Yes, please. That solution with CmdMetaParse looks better.

jameslai-dev:
Close in favor of #1176

PR #1172, opened 2026-07-27
lvm: fix device name parsing for escaped hyphens

sanket-jadhav-cse:
## Description

This pull request replaces the regex-based LVM device name parser with a left-to-right parser that correctly handles escaped hyphens in volume group and logical volume names.

## Problem

The previous implementation parsed device mapper paths using a regular expression.

While this works for many device names, it does not correctly handle all valid LVM names containing escaped hyphens (`--`). As a result, valid `/dev/mapper/...` paths can be split incorrectly, producing incorrect volume group or logical volume names.

## Solution

The parser has been rewritten to scan the device name from left to right instead of relying on a regular expression.

The new implementation:

- Treats `--` as an escaped hyphen.
- Uses the first unescaped hyphen as the separator between the volume group and logical volume names.
- Rejects malformed device names that do not contain valid volume group or logical volume components.
- Preserves the existing behavior for valid device mapper paths while correctly handling escaped hyphens.

## How I Found It

While reviewing the LVM device name parsing logic, I noticed that the parser relied entirely on a regular expression. Comparing its behavior with representative LVM device names containing escaped hyphens showed that some valid paths could be parsed incorrectly, motivating a parser that follows LVM's escaping rules directly.

## Testing

- Ran `make check`
- Verified `lvm-utils.test` passes
- Compared the parser output with `dmsetup splitname` for representative device names

### DIFF
--- a/snapper/LvmUtils.cc
+++ b/snapper/LvmUtils.cc
@@ -37,15 +37,60 @@ namespace snapper
 	pair<string, string>
 	split_device_name(const string& name)
 	{
-	    static const std::regex rx(DEV_MAPPER_DIR "/(.*[^-])-([^-].*)", std::regex::extended);
-	    std::smatch match;
+	    // Device mapper uses /dev/mapper/<vg_name>-<lv_name> format where
+	    // hyphens in names are escaped as "--" and a single "-" separates VG/LV.
+	    // Parse the name left-to-right, treating "--" as an escaped hyphen and the
+	    // first standalone "-" as the VG/LV separator.
 
-	    if (!regex_match(name, match, rx))
+	    const string prefix = DEV_MAPPER_DIR "/";
+
+	    if (!boost::starts_with(name, prefix))
 		throw std::runtime_error("failed to split device name into volume group and "
 					 "logical volume name");
 
-	    string vg_name = boost::replace_all_copy(match[1].str(), "--", "-");
-	    string lv_name = boost::replace_all_copy(match[2].str(), "--", "-");
+	    const string basename = name.substr(prefix.length());
+	    string vg_name, lv_name;
+	    size_t i = 0;
+
+	    while (i < basename.length())
+	    {
+		if (basename[i] == '-')
+		{
+		    if (i + 1 < basename.length() && basename[i + 1] == '-')
+		    {
+			vg_name += '-';
+			i += 2;
+		    }
+		    else
+		    {
+			i++;
+			break;
+		    }
+		}
+		else
+		{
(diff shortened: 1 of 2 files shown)


aschnell:
Could you provide some examples of name that so far fail? Best as tests in lvm-utils.cc.

sanket-jadhav-cse:
@aschnell Thank you for the suggestion!

I've updated the PR and added regression tests in `testsuite/lvm-utils.cc` covering the LVM device names that were not handled correctly by the previous parser.

The new test cases include:

* `/dev/mapper/vg---root`
* `/dev/mapper/vg---lv--`
* `/dev/mapper/a---b`

I also verified these examples against `dmsetup splitname` to ensure that the expected results match the behavior of the LVM tools.

Please let me know if you think additional test cases would be useful. I'm happy to add more if needed.

snapper/LvmUtils.cc
```diff
@@ -37,15 +37,60 @@ namespace snapper
...
 	{
-	    static const std::regex rx(DEV_MAPPER_DIR "/(.*[^-])-([^-].*)", std::regex::extended);
-	    std::smatch match;
+	    // Device mapper uses /dev/mapper/<vg_name>-<lv_name> format where
+	    // hyphens in names are escaped as "--" and a single "-" separates VG/LV.
+	    // Parse the name left-to-right, treating "--" as an escaped hyphen and the
+	    // first standalone "-" as the VG/LV separator.
 
-	    if (!regex_match(name, match, rx))
+	    const string prefix = string(DEV_MAPPER_DIR) + "/";
```
aschnell: You can simply use ```const string prefix = DEV_MAPPER_DIR "/";```.

snapper/LvmUtils.cc
```diff
@@ -37,15 +37,60 @@ namespace snapper
...
+
+	    if (!boost::starts_with(name, prefix))
 		throw std::runtime_error("failed to split device name into volume group and "
 					 "logical volume name");
 
-	    string vg_name = boost::replace_all_copy(match[1].str(), "--", "-");
-	    string lv_name = boost::replace_all_copy(match[2].str(), "--", "-");
+	    string remainder = name.substr(prefix.length());
```
aschnell: Can be ```const```. And maybe the name ```basename``` is more common.

aschnell:
Thanks, looks fine. Just two minor notes.

sanket-jadhav-cse:
@aschnell Thanks! I addressed both review comments and updated the code accordingly.

PR #1174, opened 2026-07-30
fix: handle CRLF line endings in AsciiFile parser

sanket-jadhav-cse:
## Summary

Strip trailing carriage returns (`\r`) when reading ASCII and gzip files.

This fixes parsing of Windows-style CRLF configuration files, where values previously retained a trailing `\r` and were parsed incorrectly.

## Changes

- Strip trailing `\r` before processing each line.
- Add a regression test covering CRLF sysconfig input.

## Testing

- Added new `sysconfig-crlf.test`.
- Existing tests continue to pass.

### DIFF
--- a/testsuite/sysconfig-crlf.txt
+++ b/testsuite/sysconfig-crlf.txt
@@ -1,0 +1,1 @@
+KEY="value"
--- a/snapper/AsciiFile.cc
+++ b/snapper/AsciiFile.cc
@@ -172,6 +172,9 @@ namespace snapper
 	if (n > 0 && buffer[n - 1] == '\n')
 	    n--;
 
+	if (n > 0 && buffer[n - 1] == '\r')
+	    n--;
+
 	line = string(buffer, 0, n);
 
 	return true;
@@ -348,6 +351,8 @@ namespace snapper
 	    if (p2)
 	    {
 		line += string(p1, p2 - p1);
+		if (!line.empty() && line.back() == '\r')
+		    line.pop_back();
 		buffer_read = p2 - buffer.data() + 1;
 		return true;
 	    }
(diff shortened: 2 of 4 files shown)


testsuite/sysconfig-crlf.txt
```diff
@@ -0,0 +1 @@
+KEY="value"
```
aschnell, line 1: Is the CR really included here? I cannot see it.

sanket-jadhav-cse: Thanks for pointing that out.

Yes, the file intentionally contains a trailing carriage return (`\r`).

I verified it locally:

$ xxd -g1 testsuite/sysconfig-crlf.txt
00000000: 4b 45 59 3d 22 76 61 6c 75 65 22 0d 0a

The trailing `0d 0a` indicates CRLF line endings.

aschnell: OK, I downloaded https://github.com/sanket-jadhav-cse/snapper/blob/fbbeea4b16768fbbd2c9fa1ba5a9931f6ee31294/testsuite/sysconfig-crlf.txt and cannot see it. I'll check after merging.

aschnell:
Thanks, look good.

aschnell:
The CR is not there after merging and git clone. Could you check? Maybe my git is configured differently.

sanket-jadhav-cse:
Thanks for checking.

I pulled the merged `master` and on my side I still see CRLF line endings in the working tree:

$ xxd -g1 testsuite/sysconfig-crlf.txt
00000000: 4b 45 59 3d 22 76 61 6c 75 65 22 0d 0a

`cat -v` also shows the trailing `^M`.

`git ls-files --eol` reports:

i/lf    w/crlf    attr/    testsuite/sysconfig-crlf.txt

So it seems our Git checkouts are handling the file differently. It may indeed be related to Git's line-ending configuration.

I've attached a screenshot of the output from my environment as well.
Screenshot 2026-07-30 182904

aschnell:
Strange. Since it has to work for me in the end I'll change the test to generate the file.

PR #1176, opened 2026-08-03
Add a wrapper to provide meta checksum and content for snbk

jameslai-dev:
This PR implements `CmdMetaParse` to provide the checksum and content of `info.xml` for
snbk, enabling the future development of the decoupled snapshot retention policy.

Detailed changes:

- `CmdMetaParse` is added to replace `CmdChecksum`. It reads the content of `info.xml` using the `cat` utility into memory during construction to minimize remote I/O. The meta checksum and content are calculated or loaded on demand via dedicated member functions.
- The OpenSSL development library is added to the project dependencies to provide the SHA256 checksum function.
- Since info.xml is not directly accessible when using SSH push, a constructor is added to
  XmlFile to load the XML content from a string.

Parent issue: #1105


client/snbk/CmdMetaParse.cc
```diff
@@ -0,0 +1,134 @@
...
+#include "../misc.h"
+
+
+namespace snapper
+{
+    CmdMetaParse::CmdMetaParse(const Shell& shell, const string& snapshot_dir,
+                               const string& cat_bin)
+        : path(snapshot_dir + "/info.xml")
+    {
+	SystemCmd::Args cmd_args = { cat_bin, path };
```
aschnell: Please add "--" between cat_bin and path.

jameslai-dev: I've added it. I did some research on the `--` option, and now I can see its importance. It can avoid a file path starting with `-` being interpreted as a command option. Thanks for your review!

client/snbk/CmdMetaParse.cc
```diff
@@ -0,0 +1,134 @@
...
+
+	content = boost::join(cmd.get_stdout(), "");
+
+	y2mil(*this);
+    }
+
+
+    string CmdMetaParse::get_checksum() const
+    {
+	if (!content.length())
```
aschnell, line 72: I'm not so sure about this. The old CmdChecksum did throw and exception which I think is better. If info.xml is empty on the source something is really broken and must be fixed. An the target the transfer should fix it in the next run. And running a cleanup algorithm with default values looks wrong.

client/snbk/CmdMetaParse.h
```diff
@@ -32,29 +34,34 @@ namespace snapper
     using std::string;
 
 
-    /**
-     * Get the checksum (e.g. sha256sum) of the file at the given path.
-     */
-    class CmdChecksum
+    struct SnapshotMeta
```
aschnell: No time entry? Isn't that needed for a cleanup.

.github/workflows/ci-tumbleweed.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         elif [ "${{ matrix.compiler }}" = "clang" ]; then
           zypper --non-interactive install --no-recommends clang
         fi
-        zypper --non-interactive install --no-recommends autoconf automake dbus-1-devel docbook-xsl-stylesheets gettext-runtime glibc-locale libacl-devel libboost_test-devel libboost_thread-devel libbtrfs-devel libbtrfsutil-devel libjson-c-devel libmount-devel libtool libxml2-devel libxslt make ncurses-devel pam-devel xz
+        zypper --non-interactive install --no-recommends autoconf automake dbus-1-devel docbook-xsl-stylesheets gettext-runtime glibc-locale libacl-devel libboost_test-devel libboost_thread-devel libbtrfs-devel libbtrfsutil-devel libjson-c-devel libmount-devel libtool libxml2-devel libxslt make ncurses-devel pam-devel xz libopenssl-devel
```
aschnell: On my Tumbleweed system libressl is installed. Not that I know the difference but using libressl here looks better.

client/snbk/CmdMetaParse.cc
jameslai-dev: Your opinion is reasonable. What about adding a `SnapshotMeta.state` attribute indicating the state of `info.xml`? If the metadata state is invalid, exclude the snapshot from cleanup algorithms.

client/snbk/CmdMetaParse.h
jameslai-dev: You are right! I did miss that part. Could I replace the usage of `TheBigThing.date` with `TheBigThing.meta.date` after adding the attribute, so that there won't be two duplicate attributes in the structure?

.github/workflows/ci-tumbleweed.yml
jameslai-dev: The `libopenssl-devel` is also listed as a build requirement in `snapper.spec.in`. Should it also be changed to `libressl`?

jameslai-dev: I'm not familiar with the two packages. It seems LibreSSL is not available from the official Debian and Ubuntu repositories. If both `libressl` and `libopenssl-devel` provide the SHA256 function, could I list them as alternatives in `snapper.spec.in`?

aschnell: AFAIS the best is to always use "pkgconfig(openssl)". The build service and zypper are supposed to understand that.

client/snbk/CmdMetaParse.h
aschnell: Yes, replacing date with meta.date looks right.

client/snbk/CmdMetaParse.cc
aschnell: Might be fine. But that has to be reported to the user so they can see fix the problem. It's not supposed to be missing/wrong in the first place.

client/snbk/CmdMetaParse.h
```diff
@@ -0,0 +1,89 @@
...
+
+	string get_checksum() const;
+	SnapshotMeta get_meta() const;
+
+	friend std::ostream& operator<<(std::ostream& s,
+	                                const CmdMetaParse& cmd_metaparse);
+
+    private:
+
+	string path;
```
aschnell: path can be const.

client/snbk/CmdMetaParse.cc
```diff
@@ -0,0 +1,153 @@
...
+	meta.state = SnapshotMeta::State::VALID;
+
+	return meta;
+    }
+
+
+    std::ostream& operator<<(std::ostream& s, const CmdMetaParse& cmd_metaparse)
+    {
+	SnapshotMeta meta = cmd_metaparse.get_meta();
+	s << "path: " << cmd_metaparse.path
```
aschnell, line 146: Also output state and date.

client/snbk/CmdMetaParse.cc
```diff
@@ -0,0 +1,153 @@
...
+	for (const xmlNode* tmp_node : getChildNodes(node, "userdata"))
+	{
+	    string key, value;
+	    getChildValue(tmp_node, "key", key);
+	    getChildValue(tmp_node, "value", value);
+	    if (!key.empty())
+		meta.userdata[key] = value;
+	}
+
+	meta.state = SnapshotMeta::State::VALID;
```
aschnell, line 137: When do we actually not get here? When the file is empty or invalid xml an exception is thrown. AFAIS if the function returns a meta it is always valid. In that case we don't need the state (which is fine for me).

jameslai-dev: Sorry, I couldn't update the descriptions immediately after pushing the changes to GitHub due to a work interruption.

If a snapshot exists on both the source and target, but the `info.xml` is broken on the target, `SnapshotMeta.date` is still populated by `probe_source`. In this scenario, the invalid state can prevent cleanup algorithms from running with the default values.

.github/workflows/ci-tumbleweed.yml
jameslai-dev: I use `pkgconfig(libcrypto)` in both `snapper.spec.in` and GitHub workflows. The GitHub workflows seem fine.
Builds for Fedora 38, 39, and Rawhide on the openSUSE Build Service failed with an unresolvable error:

```
have choice for pkgconfig(libcrypto): openssl-devel openssl3-devel
```

image

The installation seems fine in Docker containers. Is it a limitation of the build service?

client/snbk/CmdMetaParse.cc
```diff
@@ -0,0 +1,153 @@
...
+	    y2err("command '" << cmd.cmd() << "' failed: " << cmd.retcode());
+	    for (const string& tmp : cmd.get_stdout())
+		y2err(tmp);
+	    for (const string& tmp : cmd.get_stderr())
+		y2err(tmp);
+
+	    SN_THROW(Exception(_("Failed to load info.xml.")));
+	}
+
+	content = boost::join(cmd.get_stdout(), "");
```
aschnell: This could lead to invalid XML. E.g.

```
<tag attr1="val1"
attr2="val2">
```
becomes
```
<tag attr1="val1"attr2="val2">
```

So join with newlines.

client/snbk/CmdMetaParse.h
jameslai-dev: Resolved in 41fae2f

client/snbk/CmdMetaParse.cc
jameslai-dev: Resolved in a08bd81.
I use the raw format to output the date attribute. I think the raw format is sufficient for debugging and logging purposes. Using the human-readable format requires passing the UTC and ISO flags to the function. In this case, additional function parameters need to be added to `probe_source` and `probe_target`.

client/snbk/CmdMetaParse.cc
jameslai-dev: Resolved in bf07964

aschnell:
Thanks, look fine.

.github/workflows/ci-tumbleweed.yml
jameslai-dev: Resolved by setting a package preference through the OBS project config:

```
Prefer: libopenssl-devel openssl-devel
```

PR #1178, opened 2026-08-11
Remove auto-generated sysconfig-crlf.txt from the testsuite Makefile

jameslai-dev:
This PR removes the auto-generated `sysconfig-crlf.txt` file from `testsuite/Makefile.am` to prevent `make package` from depending on `make check`.

The `sysconfig-crlf.txt` is auto-generated by `testsuite/sysconfig-crlf.cc`, so I assume it does not need to be included in the Makefile.


aschnell:
Thanks for noticing.

PR #1180, opened 2026-08-21
Optimize comparisons for selected files

### LINKED ISSUE #251: snapper diff feature extremely slow even for a single file
I have a partition with millions of files which get snapshot every couple of hours.
I understand that making a diff on large trees can take ages to complete. But I do not understand why this is the case if we point to a single file.

As an (actually quite good) alternative I use meld as a GUI and can compare two snapshots in a perfect way by pointing to the proper hidden directories (works as root only)

Why is it so slow on the command line? For scripting purposes meld is not adequate.

I have from time to time the use case to revert exactly one configuration file because the app managing it happens to crash...  Over the command line it'll be a childplay but does not complete before the DBUS timeout...

I am a software developer and could help myself but I am not aware of the overall design aspects. Could you give hints for the direction to go for improving this?

EdwardBlair:
## Summary

- avoid constructing a complete snapshot-tree comparison when `undochange`, `diff`, or `xadiff` is given explicit paths
- compare only the selected paths in both library and D-Bus modes while preserving status, ignore-filter, and undo behavior
- perform D-Bus comparisons inside `snapperd`, with compatibility fallback for older daemons
- reject paths outside the configured subvolume and balance comparison/mount cleanup on success and failure

Fixes #251.

## Testing

- `make check` in a disposable openSUSE Tumbleweed Podman container
- focused `selected-files.test` coverage for modified, created, deleted, type-changed, permission-changed, ignored, duplicate, and rejected paths, plus undo execution
- `testsuite-real`: all 16 tests passed against a loopback Btrfs filesystem in a privileged disposable Podman container, including permissions, ownership, error handling, xattrs, and ACL restoration
- end-to-end A/B benchmark comparing the parent commit with this branch through the real `snapper` CLI, system D-Bus, `snapperd`, and a loopback Btrfs filesystem:
  - 100,000 baseline files
  - 20,000 unrelated endpoint differences (10,000 modified, 5,000 deleted, 5,000 created)
  - one selected file restored and correctness-checked per run
  - one warm-up followed by seven timed runs per version
  - parent median: 303.0 ms
  - optimized median: 11.3 ms
  - end-to-end speedup: 26.78x
- manual end-to-end A/B test on an Arch Linux Btrfs `/home` configuration with real snapshot history:
  - `snapper diff 1732..0 ~/.codex/config.toml`: 36.355 s with Arch's `snapper 0.13.1` / `libsnapper 8.0.0`, 0.011 s with this branch (~3,300x faster)
  - `snapper undochange 1732..0 ~/.codex/config.toml`: 21.235 s with Arch's `snapper 0.13.1` / `libsnapper 8.0.0`, 0.011 s with this branch (~1,930x faster)
  - the undo operation reported `create:0 modify:1 delete:0`, confirming that it performed a real restoration
  - timings include the CLI, D-Bus, `snapperd`, comparison, and filesystem work

I use Arch btw 😉


EdwardBlair:
Optional automated CLI/D-Bus regression coverage is available as a stacked draft PR against this PR’s head branch: https://github.com/EdwardBlair/snapper/pull/2

It runs the real `snapper` CLI and `snapperd` over an isolated D-Bus with loopback Btrfs, verifies that `diff` and `undochange` use `GetFilesForPaths` without falling back to `CreateComparison`, and checks that only the selected file is restored. I have kept it separate for now so it can be reviewed or included without expanding this PR unconditionally.

aschnell:
What I dislike about the PR is that need features are added to snapperd. In the end I cannot see a reason why this cannot be entirely be implemented in the client (with help of libsnapper).

EdwardBlair:
I see what you mean now this is the wrong layer. If I understand correctly the client should use the existing snapshot mount paths and libsnapper to compare only the explicitly requested files.. I hadn't realised this was possible because the surface-level/proxy API does not obviously expose the underlying snapshot path, which made it look as though the comparison had to stay behind the daemon boundary. I shall investigate this. It looks like it will be a much leaner change!

PR #1187, opened 2026-09-19
Document and fix common pitfalls in `default-config`

sicherha:
The default configuration in its current form gives users a false sense of security: it gives a superficial impression that reasonable measures against disk-space exhaustion are in place, but that is not the case. How would I know? Well, ... 😅

This pull request proposes the following improvements:
* Add comments to `default-config` that highlight important interdependencies and common pitfalls.
* Change the default `TIMELINE_LIMIT_*=` settings to ranges in order to facilitate graceful behaviour before disk space runs out.


aschnell:
Timeline with a lower number of 1 is also not acceptable.

sicherha:
Fair enough. So what _is_ an acceptable default?

Let me take a step back and explain. The status quo is a default configuration that _appears_ to give very reasonable safety guarantees:
* snapshots cannot occupy more than 50 % of the total disk space;
* a safety margin of 20 % is always left free.

But those two settings are ineffective unless the user manually adjusts certain other settings. I have fallen into these traps: first I ran out of disk space, then I encountered hefty performance problems caused by quota groups. I would thus like to trigger a discussion about what can be done to better protect other new users from suffering the same fate.

At the very least, I believe it would be valuable to users if the important interdependencies were mentioned right in the configuration file (proposed commit b54ba1e31c14dbcab524c8a89a95488223d673ca).

aschnell:
For me the past has shown that no configuration is possible that makes everybody happy. Either they
have no snapshot of the data they need just now or they run out of disk space. The next problem is
that people do not read the documentation.

What could be done is improve btrfs quota performance. That will AFAIU not happen. Even btrfs
developer say that performance is bad by design and it is not recommended to use it unless really
required. AFIAR there is a feature request at SUSE internally to disable quota but it is not making
good progress.

Another improvement could be that programs writing really large amounts of data to disk are
aware of btrfs snapshots and warn users before doing so. E.g. AFAIR 'zypper dup' does not so:
It assumes that when a 100 MB file gets overridden by a 110 MB file that only 10 MB additionally
will be used so.

sicherha:
I fully agree with you on both points: there can be no one-size-fits-all configuration, and users generally do not bother to study the documentation before dipping their toes in.

The question is - is it realistic to expect that of users? I would argue that software is truly user-friendly when it ...
1. provides well-balanced 'middle-of-the-road' presets,
2. anticipates common pitfalls and tries to guide users around them,
3. regards its documentation not as a prerequisite for use but as a source of in-depth information for power users.

Item 2 is what I tried to approach with commit b54ba1e31c14dbcab524c8a89a95488223d673ca: to provide relevant important information right in the usage context. (I'm of course open to any and all suggestions regarding the wording.)

My attempt at tackling item 1 was commit ee9808a36866cbd31f75bac7d570b7b217834601 - but I accept that that proposal strays too far into a direction you're not willing to go. Here I'd be interested if the openSUSE project has received concrete feedback from the field: does the default config work fine for the average long-term openSUSE user or has disk-space exhaustion turned out to be an issue? In particular, I wonder whether the longer-term (i.e. monthly and yearly) snapshots might be problematic, given how CoW snapshots inevitably grow in size over time.

aschnell:
There was feedback in the beginning and together with product management the setting were changes,
esp. the size of the root file system was increased. That was maybe 10 years ago. Since then there is not
much feedback I'm aware of.

I have added a few words to the man pages. The default configuration file is not the right place anyway
since people are not really supposed to directly edit that.

sicherha:
Isn't `default-config` used as a template that gets copied into `/etc/snapper/configs` by `snapper create-config`?

aschnell:
Yes, default-config is the template. But users are also not supposed to edit files in /etc/snapper/configs/ directly.
E.g. when a user wants to enable quota based cleanup the command is "snapper setup-quota" instead of "btrfs ..." and edit the config file.

sicherha:
> But users are also not supposed to edit files in /etc/snapper/configs/ directly.

Now that's a surprise. I fully get your point about `snapper setup-quota` - but nothing in the structure and contents of the configuration file gives the impression that it is not intended for direct human consumption. After all, it contains
* comments,
* blank lines,
* no warning at the top saying e.g. `AUTO-GENERATED; DO NOT EDIT!`.

